PR #111, opened 2024-01-29
Added CI for public docsite

dogversioning:
This PR adds a CI workflow to publish doc changes to the main cumulus docsite.

### DIFF
--- a/.github/workflows/pages.yaml
+++ b/.github/workflows/pages.yaml
@@ -1,0 +1,24 @@
+name: Update Cumulus docs
+on:
+  push:
+    branches: ["main"]
+    paths: ["docs/**"]
+
+jobs:
+  update-docs:
+    name: Update Cumulus docs
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/pages.yaml
```diff
@@ -0,0 +1,24 @@
...
+    branches: ["main"]
+    paths: ["docs/**"]
+
+jobs:
+  update-docs:
+    name: Update Cumulus docs
+    runs-on: ubuntu-latest
+    steps:
+      - name: Send workflow dispatch
+        uses: actions/github-script@v6
```
mikix, line 13: v7 is latest if you care :shrug: - every single action in the universe upgraded versions a while back to rebase on node 20

mikix (approved):

.github/workflows/pages.yaml
dogversioning: I'll change this here and also in the next library PR.

PR #117, opened 2024-06-24
Switch csv generation to QUOTE_MINIMAL

dogversioning:
Now that the dashboard supports CSV quoting, we don't need to massage commas out in this way - which is fortunate, since this is causing powerset collisions.


mikix (approved):
Oh nice

PR #122, opened 2024-10-04
Migration for new data package format

dogversioning:
- Adds a migration for data packages (and column_types)
- Adds two new items to data packages
- Some cleanup of log names, and (hopefully) an addition to prevent errors in uploading.

### DIFF
--- a/template.yaml
+++ b/template.yaml
@@ -77,11 +77,12 @@ Resources:
     Properties:
       FunctionName: !Sub 'CumulusAggFetchAuthorizer-${DeployStage}'
       Handler: src/handlers/site_upload/api_gateway_authorizer.lambda_handler
+      Layers: [arn:aws:lambda:us-east-1:336392948345:layer:AWSSDKPandas-Python311:17]
       Runtime: "python3.11"
       LoggingConfig:
         ApplicationLogLevel: !Ref LogLevel
         LogFormat: !Ref LogFormat
-        LogGroup: !Sub "/aws/lambda/CumulusAggFetchAuthorizerFunction-${DeployStage}"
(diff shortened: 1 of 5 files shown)


template.yaml
```diff
@@ -77,11 +77,12 @@ Resources:
     Properties:
       FunctionName: !Sub 'CumulusAggFetchAuthorizer-${DeployStage}'
       Handler: src/handlers/site_upload/api_gateway_authorizer.lambda_handler
+      Layers: [arn:aws:lambda:us-east-1:336392948345:layer:AWSSDKPandas-Python311:17]
```
dogversioning, line 80: Based on the error messages, I :think: this will fix the raised error, though I'm not sure why this is looking for pandas at all. But it works in the dev aggregator, so I'll have to verify this in the deployment, and may need to migrate changes back here..

dogversioning, line 80: ok, i don't think this :should: be happening, but as near as I can tell: the reason uploads (and also some other bits of the upload pipeline/dashboard apis) broke is because in python 3.11 lambdas, the isolation between lambdas is different? Layers referencing a handler which didn't use pandas were previously fine, but now it seems like, since every lambda can access the entire projects code, without this layer they fail on importing pandas?

I'll look into this and figure out if there's a better isolation pathway.

mikix: But why would an `import pandas` line even be hit in this case? Is the lambda now running a different code path because of a layer change?

That seems concerning (but maybe could happen if we now are pulling in a newer version of some pip package that added a pandas dep?).

Did we accidentally introduce an unexpected dependency on pandas ourselves and that's what's causing this? Like, do we have stacktraces for where the import comes from?

PR #123, opened 2024-10-07
Added filtering to data_packages

dogversioning:
This allows the results of the data package to be filter by either a query param (of name), or by explicit id field.

It also updates the test data to match the v2 format.

### DIFF
--- a/docs/dashboard_api.prod.yaml
+++ b/docs/dashboard_api.prod.yaml
@@ -10,6 +10,11 @@ servers:
 paths:
   /data_packages:
     get:
+      parameters:
+      - name: "name"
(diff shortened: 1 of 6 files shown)


docs/dashboard_api.prod.yaml
```diff
@@ -508,6 +513,39 @@ paths:
           content: {}
       security:
       - api_key: []
+  /data_packages/{id}:
```
dogversioning, line 516: @vlad-ignatov FYI - this should match your request, but now's a good time to spot check

mikix (approved):
Sure!

PR #124, opened 2024-10-15
Updated chart data, S3 aggregate parquet moved

dogversioning:
- Moves the `/chart_data/` endpoint to `/data_packages/[data_package_id]/chart`
- chart data now accessed via id instead of name
- updated S3 storage and glue crawlers to use data_package_id
- migration for existing aggregates
- added specific error handling for files not found in S3
- some import cleanup

### DIFF
--- a/docs/dashboard_api.prod.yaml
+++ b/docs/dashboard_api.prod.yaml
@@ -57,40 +57,6 @@ paths:
           content: {}
       security:
       - api_key: []
-  /chart-data/{subscription_name}:
-    get:
-      parameters:
-      - name: "subscription_name"
-        in: "path"
-        required: true
-        schema:
-          type: "string"
-      security:
-      - api_key: []
-    options:
-      summary: "CORS support"
-      parameters:
-      - name: "subscription_name"
-        in: "path"
-        required: true
-        schema:
-          type: "string"
-      responses:
-        "200":
-          description: "Default response for CORS method"
-          headers:
-            Access-Control-Allow-Origin:
-              schema:
-                type: "string"
-            Access-Control-Allow-Methods:
-              schema:
-                type: "string"
-            Access-Control-Allow-Headers:
-              schema:
-                type: "string"
-          content: {}
-      security:
-      - api_key: []
   /aggregates:
     get:
       security:
@@ -351,6 +317,48 @@ paths:
           content: {}
       security:
       - api_key: []
+  /data_packages/{data_package_id}/chart:
+    get:
+      parameters:
+      - name: "data_package_id"
+        in: "path"
+        required: true
+        schema:
+          type: "string"
+      - name: "column"
+        in: "query"
+        required: true
+        schema:
+          type: "string"
+      - name: "filters"
+        in: "query"
+        schema:
+          type: "string"
+      security:
+      - api_key: []
+    options:
+      parameters:
+      - name: "data_package_id"
+        in: "path"
+        required: true
+        schema:
+          type: "string"
+      responses:
+        "200":
+          description: "Default response for CORS method"
+          headers:
+            Access-Control-Allow-Origin:
+              schema:
+                type: "string"
+            Access-Control-Allow-Methods:
+              schema:
+                type: "string"
+            Access-Control-Allow-Headers:
+              schema:
+                type: "string"
+          content: {}
+      security:
+      - api_key: []
   /last_valid/{study}/{data_package}/{site}/{version}/{filename}:
     get:
       parameters:
@@ -513,39 +521,6 @@ paths:
           content: {}
       security:
       - api_key: []
-  /data_packages/{id}:
-    get:
-      parameters:
-      - name: "id"
-        in: "path"
-        required: true
-        schema:
-          type: "string"
-      security:
-      - api_key: []
-    options:
-      parameters:
-      - name: "id"
-        in: "path"
-        required: true
-        schema:
-          type: "string"
-      responses:
-        "200":
-          description: "Default response for CORS method"
-          headers:
-            Access-Control-Allow-Origin:
-              schema:
-                type: "string"
-            Access-Control-Allow-Methods:
-              schema:
-                type: "string"
-            Access-Control-Allow-Headers:
-              schema:
-                type: "string"
-          content: {}
-      security:
-      - api_key: []
   /metadata/{site}:
     get:
       parameters:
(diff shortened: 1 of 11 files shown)


src/handlers/dashboard/get_chart_data.py
```diff
@@ -22,24 +26,26 @@ def _get_table_cols(table_name: str, version: str | None = None) -> list:
     """
 
     s3_bucket_name = os.environ.get("BUCKET_NAME")
-    prefix = f"{BucketPath.CSVAGGREGATE.value}/{table_name.split('__')[0]}/{table_name}"
+    prefix = f"{enums.BucketPath.CSVAGGREGATE.value}/{dp_id.split('__')[0]}/{dp_id[:-4]}"
```
mikix, line 29: What's with ignoring the last 4 characters? Also, is there a separator we can call `rsplit` on instead? Is this meant to skip the 3-character version and the underscore? Wouldn't you need to skip 5 for that, since there are two underscores?

Also, maybe save it in a variable since you use it again below and it'd be easy to change one but not the other (plus, it lets you give it a sensible name, which might avoid questions like "what's with ignoring the last 4 characters")

src/handlers/dashboard/get_chart_data.py
```diff
@@ -91,21 +98,27 @@ def _format_payload(df: pandas.DataFrame, query_params: dict, filters: list) ->
...
+            query,
+            database=os.environ.get("GLUE_DB_NAME"),
+            s3_output=f"s3://{os.environ.get('BUCKET_NAME')}/awswrangler",
+            workgroup=os.environ.get("WORKGROUP_NAME"),
+        )
+        res = _format_payload(df, query_params, filters)
+        res = functions.http_response(200, res)
+    except errors.AggregatorS3Error:
+        res = functions.http_response(
+            404, f"Aggregate for {path_params['data_package_id']} not found"
```
mikix, line 121: nit: that's a very specific conclusion to draw from a generic "something happened error". Might be helpful to log the original error message and/or include it here too.

src/handlers/site_upload/powerset_merge.py
```diff
@@ -44,7 +44,7 @@ def __init__(self, event):
         self.study = s3_key_array[1]
         self.data_package = s3_key_array[2].split("__")[1]
         self.site = s3_key_array[3]
-        self.version = s3_key_array[4]
+        self.version = s3_key_array[4][-3:]
```
mikix, line 47: Can we break on a `split` call or something? Assuming the length feels brittle.

tests/site_upload/test_powerset_merge.py
```diff
@@ -92,17 +92,17 @@
...
+        #     # merged by substr match
+        #     "./tests/test_data/count_synthea_patient.parquet",
+        #     f"/{EXISTING_STUDY}/{EXISTING_STUDY}__{EXISTING_DATA_P[0:-2]}/{EXISTING_SITE}/"
+        #     f"{EXISTING_STUDY}__{EXISTING_DATA_P[0:-2]}_{EXISTING_VERSION}/encount.parquet",
+        #     f"/{EXISTING_STUDY}/{EXISTING_STUDY}__{EXISTING_DATA_P[0:-2]}/"
+        #     f"{EXISTING_STUDY}__{EXISTING_DATA_P[0:-2]}_{EXISTING_VERSION}/encount.parquet",
+        #     False,
+        #     200,
+        #     ITEM_COUNT + 4,
+        # ),
```
mikix, line 105: Should this be deleted or changed?

tests/site_upload/test_powerset_merge.py
```diff
@@ -136,6 +136,7 @@ def test_powerset_merge_single_upload(
     mock_notification,
 ):
     s3_client = boto3.client("s3", region_name="us-east-1")
+    s3_res = s3_client.list_objects_v2(Bucket=TEST_BUCKET)
```
mikix, line 139: :scissors: ?

mikix (approved):

docs/dashboard_api.prod.yaml
```diff
@@ -351,6 +317,48 @@ paths:
           content: {}
       security:
       - api_key: []
+  /data_packages/{data_package_id}/chart:
```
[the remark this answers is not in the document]
dogversioning: hyphens are for static urls - underscores in brackets are variable names in python bound to whatever appears in that position in the url.

docs/dashboard_api.prod.yaml
```diff
@@ -580,12 +577,23 @@ paths:
           content: {}
       security:
       - api_key: []
-  /study-periods:
+  /data_packages/{data_package_id}:
```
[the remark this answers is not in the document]
dogversioning: This is autogenerated from API gateway, I have no control over it.

src/handlers/dashboard/get_chart_data.py
dogversioning: hm - yeah, test data artifact, will bump out to five or flip to looking for last dunder.

src/handlers/dashboard/get_chart_data.py
dogversioning: The error should get logged to cloudwatch, security asked us to have no stack traces in responses and to generally have these error messages be unhelpful while the API is externally accessible.

src/handlers/site_upload/powerset_merge.py
dogversioning: yeah, same as above, will change.

tests/site_upload/test_powerset_merge.py
dogversioning: yeah lemme go back and uncomment this.

docs/dashboard_api.prod.yaml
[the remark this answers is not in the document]
dogversioning: ah wait i see what you're saying - will fix.

scripts/migrations/migration.004.s3_name_with_id.py
```diff
@@ -0,0 +1,48 @@
...
+def s3_name_with_id(bucket: str):
+    client = boto3.client("s3")
+    res = client.list_objects_v2(Bucket=bucket)
+    contents = res["Contents"]
+    moved_files = 0
+    for s3_file in contents:
+        if s3_file["Key"].split("/")[0] == "aggregates":
+            key = s3_file["Key"]
+            key_array = key.split("/")
+            if len(key_array[3]) == 3:
```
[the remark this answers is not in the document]
dogversioning: I don't want to do this because i can't guarantee that this will be a valid array position for keys from other folders.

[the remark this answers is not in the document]
mikix: That's why you put it after the len() check: `if len(key_array[3]) == 3 and key_array[0] == "aggregates":`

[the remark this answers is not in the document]
mikix: Oh right - i thought that len check was on the array. Well then you put the key_array[0] bit first:

`if key_array[0] == "aggregates" and len(key_array[3]) == 3:`

Point is, you grab "Key" and split it twice over two if checks - just looked like room to simplify. But this is also disposable migration code.

src/handlers/dashboard/get_chart_data.py
mikix: Maybe add a comment to that effect? Like... if a third developer comes across this in 2 years, they won't have that context. Something like "# don't be too informative about internal state in our error messages" (but in Matt voice) above this generic response.

dogversioning: ok, added a brief note.

PR #127, opened 2024-10-15
Cast to varchar during coalesce

dogversioning:
This prevents date types from interferring with the COALESCE check in the chart data function checking for null rows.

### DIFF
--- a/src/handlers/dashboard/get_chart_data.py
+++ b/src/handlers/dashboard/get_chart_data.py
@@ -62,7 +62,10 @@ def _build_query(query_params: dict, filters: list, path_params: dict) -> str:
         group_str = f"{query_params['stratifier']}, {group_str}"
         columns.remove(query_params["stratifier"])
     if len(columns) > 0:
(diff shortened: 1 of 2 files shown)


src/handlers/dashboard/get_chart_data.py
```diff
@@ -62,7 +62,10 @@ def _build_query(query_params: dict, filters: list, path_params: dict) -> str:
         group_str = f"{query_params['stratifier']}, {group_str}"
         columns.remove(query_params["stratifier"])
     if len(columns) > 0:
-        coalesce_str = f"WHERE COALESCE ({','.join(columns)}) IS NOT Null AND"
+        coalesce_str = (
+            f"WHERE COALESCE (cast({' AS VARCHAR), cast('.join(columns)} AS VARCHAR)) "
+            "IS NOT Null AND"
```
mikix, line 67: nit: unusual casing on `Null`

mikix (approved):

src/handlers/dashboard/get_chart_data.py
dogversioning: why on earth did i do that

PR #129, opened 2024-10-16
Updated stratifier behavior

dogversioning:
- Fixes stratifier to prevent returning all data
- Adds a counts section to stratifier to match new vintage of chart data API

### DIFF
--- a/src/handlers/dashboard/get_chart_data.py
+++ b/src/handlers/dashboard/get_chart_data.py
@@ -17,7 +17,7 @@ logger = logging.getLogger()
 logger.setLevel(log_level)
 
 
-def _get_table_cols(dp_id: str, version: str | None = None) -> list:
+def _get_table_cols(dp_id: str) -> list:
     """Returns the columns associated with a table.
 
     Since running an athena query takes a decent amount of time due to queueing
@@ -26,11 +26,11 @@ def _get_table_cols(dp_id: str, version: str | None = None) -> list:
     """
 
     s3_bucket_name = os.environ.get("BUCKET_NAME")
-    dp_name = dp_id.rsplit("__", 1)[0]
-    prefix = f"{enums.BucketPath.CSVAGGREGATE.value}/{dp_id.split('__')[0]}/{dp_name}"
+    study, name, version = dp_id.split("__")
+    prefix = f"{enums.BucketPath.CSVAGGREGATE.value}/{study}/{study}__{name}"
     if version is None:
         version = functions.get_latest_data_package_version(s3_bucket_name, prefix)
-    s3_key = f"{prefix}/{version}/{dp_name}__aggregate.csv"
+    s3_key = f"{prefix}/{version}/{study}__{name}__aggregate.csv"
     s3_client = boto3.client("s3")
     try:
         s3_iter = s3_client.get_object(
@@ -98,14 +98,16 @@ def _format_payload(
     if "stratifier" in query_params.keys():
         payload["stratifier"] = query_params["stratifier"]
         counts = {}
-        for unique_val in df[query_params["column"]]:
-            df_slice = df[df[query_params["column"]] == unique_val]
+        for unique_val in df[query_params["column"]].unique():
+            column_mask = df[query_params["column"]] == unique_val
+            df_slice = df[column_mask]
             df_slice = df_slice.drop(columns=[query_params["stratifier"], query_params["column"]])
             counts[unique_val] = int(df_slice[count_col].sum())
         payload["counts"] = counts
         data = []
         for unique_strat in df[query_params["stratifier"]].unique():
-            df_slice = df[df[query_params["stratifier"]] == unique_strat]
+            strat_mask = df[query_params["stratifier"]] == unique_strat
+            df_slice = df[strat_mask]
             df_slice = df_slice.drop(columns=[query_params["stratifier"]])
             rows = df_slice.values.tolist()
             data.append({"stratifier": unique_strat, "rows": rows})
--- a/src/handlers/dashboard/get_chart_data.py
+++ b/src/handlers/dashboard/get_chart_data.py
@@ -48,10 +48,11 @@ def _build_query(query_params: dict, filters: list, path_params: dict) -> str:
     columns = _get_table_cols(dp_id)
     filter_str = filter_config.get_filter_string(filters)
     if filter_str != "":
-        filter_str = f"AND {filter_str}"
+        filter_str = f"AND {filter_str} "
     count_col = next(c for c in columns if c.startswith("cnt"))
     columns.remove(count_col)
     select_str = f"{query_params['column']}, sum({count_col}) as {count_col}"
+    strat_str = ""
     group_str = f"{query_params['column']}"
     # the 'if in' check is meant to handle the case where the selected column is also
     # present in the filter logic and has already been removed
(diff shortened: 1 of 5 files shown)


src/handlers/dashboard/get_chart_data.py
```diff
@@ -61,29 +62,33 @@ def _build_query(query_params: dict, filters: list, path_params: dict) -> str:
         select_str = f"{query_params['stratifier']}, {select_str}"
         group_str = f"{query_params['stratifier']}, {group_str}"
         columns.remove(query_params["stratifier"])
+        strat_str = f'AND {query_params["stratifier"]} IS NOT NULL '
```
mikix, line 65: I'm not seeing any quoting of these query params? We should probably be escaping them, then adding double quotes (or verifying that they match `[a-zA-Z]+`), yeah? That's what the ruff warning `S608` is about, which is quieted below, yeah.

Does not need to block this PR, but feels important.

src/handlers/dashboard/get_chart_data.py
```diff
@@ -92,13 +97,20 @@ def _format_payload(df: pandas.DataFrame, query_params: dict, filters: list) ->
     payload["totalCount"] = int(df["cnt"].sum())
     if "stratifier" in query_params.keys():
         payload["stratifier"] = query_params["stratifier"]
+        counts = {}
+        for unique_val in df[query_params["column"]]:
+            df_slice = df[df[query_params["column"]] == unique_val]
```
mikix, line 102: I'm a poor man from a poor family - what is this code doing? You grab the column slice - then compare that to a variable and look up a boolean in the dataframe? Or is this some Pandas bullshit where it overrides `==` to return a special object that can be sliced like a lambda by the `[]` operator (and presumably for normal cases just be truthy)?

So this is getting all rows where that column matches the value. Got it. I'm ambivalent on whether this is more or less readable than whatever method-based version Pandas offers to do the same thing - just threw me for a bit.

src/handlers/dashboard/get_chart_data.py
```diff
@@ -92,13 +97,20 @@ def _format_payload(df: pandas.DataFrame, query_params: dict, filters: list) ->
     payload["totalCount"] = int(df["cnt"].sum())
     if "stratifier" in query_params.keys():
         payload["stratifier"] = query_params["stratifier"]
+        counts = {}
+        for unique_val in df[query_params["column"]]:
```
mikix, line 101: Why does the stratifier loop below use `.unique()` but this doesn't?

src/handlers/dashboard/get_chart_data.py
```diff
@@ -112,17 +124,19 @@ def chart_data_handler(event, context):
     del context
     query_params = event["queryStringParameters"]
     filters = event["multiValueQueryStringParameters"].get("filter", [])
+    if "filter" in query_params and filters == []:
+        filters = [query_params["filter"]]
```
mikix, line 128: Why the `filters == []` check (which could be `not filters` I think)? Do we want to ignore a `filter` query entirely if `multiValueQueryStringParameters` is set?

tests/dashboard/test_get_chart_data.py
```diff
@@ -99,7 +101,9 @@ def test_build_query(query_params, filters, path_params, query_str):
 )
 def test_format_payload(query_params, filters, expected_payload):
     df = mock_data_frame(filters)
-    payload = get_chart_data._format_payload(df, query_params, filters)
+    payload = get_chart_data._format_payload(df, query_params, filters, "cnt")
+    print(query_params)
+    print(payload)
```
mikix, line 106: :scissors: ?

tests/dashboard/test_get_chart_data.py
```diff
@@ -113,11 +117,14 @@ def test_get_data_cols(mock_bucket):
...
     lambda query_params, filters, path_params: (
-        "SELECT gender, sum(cnt) as cnt"
-        f'FROM "{TEST_GLUE_DB}"."test_study" '
-        "WHERE COALESCE (race) IS NOT Null AND gender IS NOT Null "
-        "AND gender LIKE 'female' "
-        "GROUP BY gender",
+        (
+            "SELECT gender, sum(cnt) as cnt"
+            f'FROM "{TEST_GLUE_DB}"."test_study" '
+            "WHERE COALESCE (race) IS NOT Null AND gender IS NOT Null "
```
mikix, line 123: There's that odd `Null` again

mikix (approved):

src/handlers/dashboard/get_chart_data.py
dogversioning: >  Or is this some Pandas bullshit 

Ding ding ding. `df[query_params["column"]] == unique_val` creates a Series of true/false values, and `df[]` around that takes that input and uses it as a mask to select only the rows that are true. I could explictly store this thing in a variable with mask in the name?

src/handlers/dashboard/get_chart_data.py
dogversioning: in experimenting with this - I can't explictly say in cloudformation 'this parameter can be an array', so i'm not sure if it's going to be in `queryStringParameters` or `multiValueQueryStringParameters` in the AWS inbound event to the lambda. so I need to check both.

tests/dashboard/test_get_chart_data.py
dogversioning: yep

tests/dashboard/test_get_chart_data.py
dogversioning: I think this is autocompleteing itself in someplace

src/handlers/dashboard/get_chart_data.py
dogversioning: the stratifier label is a one-> many relationship to the column, which in practice means it shows up a bunch of time in the dataframe.

src/handlers/dashboard/get_chart_data.py
dogversioning: yeah i was just thinking about how gross this approach to SQL construction was - I may make a ticket for this rather than deal with it as part of the PR

src/handlers/dashboard/get_chart_data.py
mikix: Should we throw a .unique() on here too though?

dogversioning: no, because here we're summing on the column values, which gives us a single output row.

src/handlers/dashboard/get_chart_data.py
dogversioning: if we got both (which I don't think should happen), `queryStringParameters` would have to be a subset of `multiValueQueryStringParameters`

src/handlers/dashboard/get_chart_data.py
mikix: Yeah I guess you wouldn't get the wrong answer, you'd just re-calculate it multiple times.

dogversioning: ah! i see what you're saying now, sorry

src/handlers/dashboard/get_chart_data.py
dogversioning: moved these to mask vars for readability

src/handlers/dashboard/get_chart_data.py
dogversioning: #130

Issue #116, opened 2024-04-24
Change auth to use AWS secrets for storage


dogversioning:
We want a schema something like the following:

- A user running upload at a site gets an API key
- Two API keys are kept in secrets - on initial create, one is the provided api key
- These keys are rotated, one becoming the 'old' key and generating a new key, on a periodic basis
- When a user sends us the old key, we should pass back the new key, and the library should do :something: with that to update for the next call.

PR #137, opened 2024-10-28
Isolated lambdas, moved auth to secrets manager

dogversioning:
- Lambda reorg
  - Seperated all the lambdas into their own subfolders, and joined in the shared folder via symlink - this enable them to all be deployed seperately, which speeds up deployment in 90% of cases by a good clip
  - Removed the extra folder layer near the top of the tree
  - Updated cloudformation entrypoint to avoid reuploading the entire repo each time
  - did some import cleanup while i was in touching everything
- Moved user logins to secrets manager (#116)

There's a lot of files in this PR, but a good 90% of it is just renaming and adding symbolic links.


.pre-commit-config.yaml
```diff
@@ -1,10 +1,11 @@
 default_install_hook_types: [pre-commit, pre-push]
 repos:
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.2.1
+    rev: v0.5.7
```
dogversioning, line 4: Noting for posterity that ruff is now up to v0.7.1 - but here i'm just syncing this to this repo's version

scripts/credential_management.py
```diff
@@ -26,26 +26,10 @@ def _put_s3_data(name: str, bucket_name: str, client, data: dict, path: str = "a
     client.upload_fileobj(Bucket=bucket_name, Key=f"{path}/{name}", Fileobj=b_data)
 
 
-def create_auth(client, bucket_name: str, user: str, auth: str, site: str) -> str:
+def create_auth(client, user: str, auth: str, site: str) -> str:
```
dogversioning, line 29: this is now just a convenience script for getting the password hash

src/site_upload/api_gateway_authorizer/api_gateway_authorizer.py
```diff
@@ -4,23 +4,42 @@
...
+    return json.loads(get_secret_value_response["SecretString"])
+
+
 def lambda_handler(event, context):
     del context
     # ---- aggregator specific logic
-    user_db = get_s3_json_as_dict(
-        os.environ.get("BUCKET_NAME"), f"{BucketPath.ADMIN.value}/auth.json"
-    )
+    user_db = get_secret()
```
dogversioning, line 42: this is section that is functionality change related to secrets.

src/site_upload/api_gateway_authorizer/api_gateway_authorizer.py
```diff
@@ -4,23 +4,42 @@
...
+    secret_name = os.environ.get("SECRET_NAME")
+    region_name = os.environ.get("REGION")
+
+    session = boto3.session.Session()
+    client = session.client(service_name="secretsmanager", region_name=region_name)
+
+    try:
+        get_secret_value_response = client.get_secret_value(SecretId=secret_name)
+    except ClientError as e:
+        raise e
```
mikix, line 34: why this try/except block, if you're not doing anything with the error?

scripts/credential_management.py
```diff
@@ -110,19 +94,11 @@ def delete_meta(client, bucket_name: str, site: str) -> bool:
...
             args.create_auth[2],
         )
-        print(f"{id_str} created")
-    elif args.delete_auth:
-        succeeded = delete_auth(s3_client, bucket, args.delete_auth)
-        if succeeded:
-            print(f"Removed {args.delete_auth}")
-        else:
-            print(f"{args.delete_auth} not found")
+        print(id_str)
```
mikix, line 101: :scissors:? Or modify into something more user-oriented, like you had before with `print(f"{id_str} created")`

mikix (approved):

src/site_upload/api_gateway_authorizer/api_gateway_authorizer.py
dogversioning: this was the stock 'get the secrets' behavior, which i left in here mostly because this file is stock AWS code with a lil seasoning on top. but i could trim this.

scripts/credential_management.py
dogversioning: so right now this isn't actually doing any interaction with secrets manager - so this is basically outputting the key/value pair you'd add to the secrets dict. I could suggest something like that as output text.

PR #138, opened 2024-10-29
Type casting on date filter fix

dogversioning:
- Updates how we try to cast to dates from strings in the chart data endpoint
- Light cleanup on credentials from previous commit


scripts/credential_management.py
```diff
@@ -29,7 +29,7 @@ def _put_s3_data(name: str, bucket_name: str, client, data: dict, path: str = "a
 def create_auth(client, user: str, auth: str, site: str) -> str:
     """Adds a new entry to the auth dict used to issue pre-signed URLs"""
     site_id = _basic_auth_str(user, auth).split(" ")[1]
-    return f'"{site_id}"": {{"site":{site}}}'
+    return f'"{site_id}"": {{"user": {user}, "site":{site}}}'
```
dogversioning, line 32: As soon as this became a hand managed process, having a username in here just for IDing the record became very valuable

mikix (approved):

PR #139, opened 2024-11-04
Refactor of get chart data endpoint

dogversioning:


src/dashboard/get_chart_data/templates/get_chart_data.sql.jinja
```diff
@@ -0,0 +1,65 @@
...
+        {%- endif  %}
+        {%- if stratifier_column %}
+        AND "{{ stratifier_column }}" IS NOT NULL
+        {%- endif  %}
+    GROUP BY
+        {%- if stratifier_column %}
+        "{{ stratifier_column }}", "{{ data_column }}"
+        {%- else %}
+        "{{ data_column }}"
+        {%- endif %}
```
mikix, line 37: I will point out here that because you don't care about the prettiness of the resulting SQL (right?), you can instead try to make the jinja code more readable / maintainable.

Things like reducing the duplication here:
```
        {%- if stratifier_column %}
        "{{ stratifier_column }}",
        {%- endif %}
        "{{ data_column }}"
```

And/or proper indenting throughout for the if/else blocks, plus whatever else. The fact that the jinja in Library is usually hard to read is _partly_ our fault.

src/dashboard/get_chart_data/templates/get_chart_data.sql.jinja
```diff
@@ -0,0 +1,65 @@
...
+)
+SELECT * FROM non_null
+{%- if numeric.has_filter(filter_configs) == 'True' %}
+WHERE {{ numeric.get_filters(filter_configs) }}
+{%- endif %}
+UNION
+SELECT * from has_null
+{%- if numeric.has_filter(filter_configs) == 'True' %}
+WHERE {{ numeric.get_filters(filter_configs) }}
+{%- endif %}
```
mikix, line 59: Am I missing something? This looks the same as the non-null block above. Why separate out the nulls and non-nulls into two selects?

src/dashboard/get_chart_data/templates/filter_numeric.sql.jinja
```diff
@@ -0,0 +1,48 @@
...
+{%- elif filter_type == 'lt'-%}
+{{ data }} < {{ bound }}
+{%- elif filter_type == 'lte'-%}
+{{ data }} <= {{ bound }}
+{%- else -%}
+    not found {{ filter_type }}
+{%- endif -%}
+{%- endmacro -%}
+
+{%- macro is_filter_in_list(filter_configs) -%}
```
mikix, line 24: OK so this is more about validating that the incoming filters are well-formed/valid rather than checking that they meet some kind of condition. Maybe `is_filter_valid` might be more immediately clear?

src/dashboard/get_chart_data/templates/filter_numeric.sql.jinja
```diff
@@ -0,0 +1,48 @@
...
+    {%- for f in filter_configs -%}
+        {%- if f.filter_type in filters -%}
+            {%- set ns.matches = ns.matches + [f] -%}
+        {%- endif -%}
+    {% endfor -%}
+{%- endmacro -%}
+
+{%- macro get_filters(filter_configs) -%}
+    {{ is_filter_in_list(filter_configs) }}
```
mikix, line 33: Jinja is a cursed language. I understand why you did the namespace games here. But maybe throw some comments around the place explaining that jinja does not have the ability to do actual pure-logic methods / modify variables out of scope.

But also... couldn't you just do this filtering on the python side? And then avoid all this in jinja?

src/dashboard/get_chart_data/templates/filter_numeric.sql.jinja
```diff
@@ -0,0 +1,48 @@
+{%- import 'syntax.sql.jinja' as syntax -%}
+
+{%- set ns =namespace(matches=[]) -%}
+{%- set filters=['eq', 'ne', 'gt', 'gte', 'lt', 'lte'] -%}
+
+{%- macro render_filter( data, filter_type, bound) -%}
+{%- if filter_type == 'eq'-%}
+{{ data }} = {{ bound }}
```
mikix, line 8: Add double quotes around these column references?

src/dashboard/get_chart_data/templates/filter_numeric.sql.jinja
```diff
@@ -0,0 +1,48 @@
...
+
+{%- macro get_filters(filter_configs) -%}
+    {{ is_filter_in_list(filter_configs) }}
+    {%- for config in ns.matches %}
+        {{ syntax.and_delineate(loop) }}{{ render_filter(config['data'],config['filter_type'],config['bound']) }}
+    {%- endfor -%}
+    {%- set ns.matches = [] -%}
+{%- endmacro -%}
+
+{%- macro has_filter(filter_configs) %}
```
mikix, line 40: suggestion nit: you could drop this method if you put the WHERE inside get_filters and checked the count there... :balancing hands motion:  Less flexible overall, but slightly cleaner to have the code be self-contained like that.

src/dashboard/get_chart_data/templates/filter_numeric.sql.jinja
dogversioning: yeah, commenting will come at the end.

i :could: do the splitting on the python side - i'm then maintaining the list of filters in both jinja and python, which i don't love, but it would avoid this whole thing.

src/dashboard/get_chart_data/templates/filter_numeric.sql.jinja
dogversioning: this gets wonky with the multiple filter types above - my instinct was to do some sort of `string.has_filter or date.has_filter` kind of logic later on.

src/dashboard/get_chart_data/templates/get_chart_data.sql.jinja
mikix: > yeah i think we can ~~make~~ **ditch** some compromises over here that we cant/wont in the library.

Fixed that for ya :smile:

tests/dashboard/test_filter_inline.py
dogversioning, line 1: this is mostly a diff from test_filter_config - same inputs, different outputs.

src/dashboard/get_chart_data/templates/filter_numeric.sql.jinja
dogversioning: ok i ended up doing some rework to move more of this logic into python.

src/dashboard/get_chart_data/templates/filter_numeric.sql.jinja
dogversioning: Removed this file, but generally did passes to ensure fields were properly quoted

tests/dashboard/test_get_chart_data.py
```diff
@@ -29,53 +29,230 @@ def mock_data_frame(filter_param):
...
+    "query_params,filter_groups,path_params,query_str",
     [
         (
             {"column": "gender"},
             [],
             {"data_package_id": "test_study"},
-            f'SELECT gender, sum(cnt) as cnt FROM "{TEST_GLUE_DB}"."test_study" '
-            "WHERE COALESCE (cast(race AS VARCHAR)) IS NOT NULL AND gender IS NOT NULL "
-            "GROUP BY gender ORDER BY gender",
+            f"""SELECT
```
dogversioning, line 38: these are getting particularly irritating to maintain - I did open #140 for considering ditching this in favor of an actual in-DB test.

mikix (approved):

PR #141, opened 2024-11-11
Chart data cleanup

dogversioning:
This PR addresses some edge cases found in dashboard testing:

- String comparisons can be used on number, so casting is required
- Incorrect quoting on date columns
- Fixed detection of nulls in incoming filter array objects
- `matches` action name was incorrect

### DIFF
--- a/src/dashboard/get_chart_data/get_chart_data.py
+++ b/src/dashboard/get_chart_data/get_chart_data.py
@@ -29,7 +29,7 @@ INLINE_FILTERS = (
     "strContainsCI",
     "strStartsWithCI",
     "strEndsWithCI",
-    "strMatchesCI",
+    "matchesCI",
     "strNotEq",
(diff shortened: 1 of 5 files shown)


src/dashboard/get_chart_data/get_chart_data.py
```diff
@@ -165,9 +165,10 @@ def _build_query(query_params: dict, filter_groups: list, path_params: dict) ->
                 raise errors.AggregatorFilterError(
                     f"Invalid filter type {filter_config[1]} requested."
                 )
-
-        inline_configs.append(config_params)
-        none_configs.append(none_params)
+        if config_params != []:
```
mikix, line 168: `if config_params:` might be more natural, unless there's a reason you want to append a `None` value.

mikix (approved):

Issue #125, opened 2024-10-15
Crawl files on upload


dogversioning:
If we need data to be available from sites same day, we may want to change the way we do crawling by default.

We could, on a new aggregate type showing up for the first time, create a crawler targeted at that specific file, run it, and then remove the crawler. [This post](https://medium.com/@Lux_real/how-to-create-aws-glue-crawler-for-specific-files-based-on-a-pattern-2926c5df65c6) outlines some of the general idea.

Issue #70, opened 2023-04-13
technical debt: renaming to limit use of the term 'metadata'


dogversioning:
we now have too many things that are being referred to as metadata, and it is making things a bit fussy to message/manage, so we should do some light renaming.

items to touch:
- read/write metadata -> read/write s3_json
- site metadata -> data packaging status?
- bucket names should not contain `meta`
- study medata :might: be ok.

Once this is done, there may be some redundancies in shared functions that can be trimmed out - perhaps one central 'write json to s3' function anything can use?

PR #142, opened 2024-12-03
Added flat file handling

dogversioning:
This makes the following changes:
- the S3Manager class is now decoupled from powerset_merge, and is in the shared folder (and intended to be used by alternate processing methods going forward)
- A new processing type, `process_flat`, moves flat files into new location types in S3
  - Some changes to `process_upload` to shuttle things off to different queues


src/shared/enums.py
```diff
@@ -33,9 +35,19 @@ class JsonFilename(enum.Enum):
     COLUMN_TYPES = "column_types"
     TRANSACTIONS = "transactions"
     DATA_PACKAGES = "data_packages"
+    FLAT_PACKAGES = "flat_packages"
     STUDY_PERIODS = "study_periods"
 
 
+class StudyPeriodMetadataKeys(enum.Enum):
```
dogversioning, line 42: changes from here down in this file are just alpha sorting.

src/shared/functions.py
```diff
@@ -158,7 +158,7 @@ def write_metadata(
     s3_client.put_object(
         Bucket=s3_bucket_name,
         Key=f"{enums.BucketPath.META.value}/{meta_type}.json",
-        Body=json.dumps(metadata, default=str),
+        Body=json.dumps(metadata, default=str, indent=2),
```
dogversioning, line 161: unrelated, but since i was manually pretty-ifying json a ton when downloading data from S3, I decided to just write the files out this way.

src/shared/s3_manager.py
```diff
@@ -0,0 +1,235 @@
...
+            f"{self.study}__{self.data_package}__aggregate.parquet"
+        )
+        self.csv_aggregate_path = (
+            f"s3://{self.s3_bucket_name}/{enums.BucketPath.CSVAGGREGATE.value}/"
+            f"{self.study}/{self.study}__{self.data_package}/"
+            f"{self.version}/"
+            f"{self.study}__{self.data_package}__aggregate.csv"
+        )
+        # TODO: Taking out a folder layer to match the depth of non-site aggregates
+        # Revisit when targeted crawling is implemented
```
dogversioning, line 63: To expand on this a bit here: [this bit](https://github.com/smart-on-fhir/cumulus-aggregator/blob/7e72e89205a0e25afd304cff7f90139e78bad223/template.yaml#L771) of the glue config specifies that table names will be taken from the fifth directory down. Since we're introducing site as a concept here, adding a new layer, I elected to strip one element out of the "ideal" path. 

#125 discusses a way around this, which basically amounts to creating a crawler on demand and pointing it at just the directory you want to crawl new data in.

src/site_upload/powerset_merge/powerset_merge.py
dogversioning, line 1: I didn't want to touch this too much since this is already pretty unwieldy, but this module maybe should be renamed 'process_powersets' or 'process_cubes'? I also need to do a pass and see if there's more stuff in here that can/should get moved to the manager, now that that's seperate.

tests/conftest.py
```diff
@@ -65,7 +65,7 @@ def _init_mock_data(s3_client, bucket, study, data_package, version):
     )
 
 
-@pytest.fixture(autouse=True)
+@pytest.fixture(scope="session", autouse=True)
```
dogversioning, line 68: one wierd trick to eliminate all those OS env var imports, coming to other repos near you shortly.

src/shared/awswrangler_functions.py
```diff
@@ -35,3 +38,25 @@ def get_s3_study_meta_list(
...
+        f"s3://{bucket_name}/{bucket_root}" f"/{subbucket_path}"
+    )
+    last_valid_df = last_valid_df.apply(lambda x: x.strip() if isinstance(x, str) else x).replace(
+        '""', numpy.nan
+    )
+    awswrangler.s3.to_csv(
+        last_valid_df,
+        to_path,
+        index=False,
+        quoting=csv.QUOTE_MINIMAL,
```
mikix, line 61: Minimal is my own preferred dialect, but if this csv file is headed towards the dashboard, the way Vlad has talked about quoting makes me think he expects CSVs to be using the `csv.QUOTE_STRINGS` dialect (or at least `csv.QUOTE_NONNUMERIC` if you don't have Python 3.12).

src/shared/functions.py
```diff
@@ -182,6 +182,22 @@ def move_s3_file(s3_client, s3_bucket_name: str, old_key: str, new_key: str) ->
         raise S3UploadError
 
 
+def get_s3_keys(s3_client, s3_bucket_name: str, prefix: str, token: str | None = None) -> list:
```
mikix, line 185: nit: treat yourself to a return type of `list[str]`

src/shared/functions.py
```diff
@@ -182,6 +182,22 @@ def move_s3_file(s3_client, s3_bucket_name: str, old_key: str, new_key: str) ->
...
+    """Gets the list of all keys in S3 starting with the prefix"""
+    if token:
+        res = s3_client.list_objects_v2(
+            Bucket=s3_bucket_name, Prefix=prefix, ContinuationToken=token
+        )
+    else:
+        res = s3_client.list_objects_v2(Bucket=s3_bucket_name, Prefix=prefix)
```
mikix, line 192: nit: couldn't you just unconditionally pass along `ContinuationToken=token`? I assume that `list_objects_v2` has a default of `None` there, such that these two paths are equal? Did not read docs though :smile:

src/shared/s3_manager.py
```diff
@@ -0,0 +1,235 @@
...
+        # We are excluding COLUMN_TYPES explicitly from this first check because,
+        # by design, it should never have a site field in it - the column types
+        # are tied to the study version, not a specific site's data
+        if extra_items is None:
+            extra_items = {}
+        if site is None and meta_type != enums.JsonFilename.COLUMN_TYPES.value:
+            site = self.site
+        if metadata is None:
+            metadata = self.metadata
+        metadata = functions.update_metadata(
```
mikix, line 210: this local variable `metadata` is never used

src/site_upload/process_flat/process_flat.py
```diff
@@ -0,0 +1,70 @@
...
+            "type": "flat",
+        },
+    )
+    manager.update_local_metadata(enums.TransactionKeys.LAST_DATA_UPDATE.value, site=manager.site)
+    manager.update_local_metadata(
+        enums.ColumnTypesKeys.LAST_DATA_UPDATE.value,
+        value=column_dict,
+        metadata=manager.types_metadata,
+        meta_type=enums.JsonFilename.COLUMN_TYPES.value,
+    )
```
mikix, line 54: This is called twice with the same key and different values - can these be combined? (I might not be understanding how local metadata works)

src/site_upload/process_flat/process_flat.py
```diff
@@ -0,0 +1,70 @@
...
+    manager.write_local_metadata(
+        metadata=manager.types_metadata, meta_type=enums.JsonFilename.COLUMN_TYPES.value
+    )
+    manager.write_local_metadata()
```
mikix, line 58: Similar question here - the second call I assume ends up overwriting the first?

src/site_upload/process_upload/process_upload.py
```diff
@@ -23,15 +23,35 @@ def process_upload(s3_client, sns_client, s3_bucket_name: str, s3_key: str) -> N
...
     path_params = s3_key.split("/")
     study = path_params[1]
-    data_package = path_params[2]
+    # This happens when we're processing flat files, due to having to condense one
+    # folder layer.
+    # TODO: revisit on targeted crawling
+    if "__" in path_params[2]:
+        data_package = path_params[2].split("__")[1]
+    else:
+        data_package = path_params[2]
```
mikix, line 32: nit: I feel like this repo is due for a AggPath class or similar that abstracts all of the path assumptions / evolutions over time. There's a lot of path[2] and splitting going on across the repo, instead of in one nice central place.

mikix (approved):

src/shared/awswrangler_functions.py
dogversioning: I swear we had a contract at one point indicating this format was ok.

src/shared/functions.py
dogversioning: I tried that, and no, unfortunately.

src/site_upload/process_flat/process_flat.py
dogversioning: this is one of those 'i made this choice for backwards compatibility at the time' issues that has caused downstream boondoggles - `meta_type` determines which of the two cached metadata dicts (transaction details, column types) this is written to. So these actually are writing to completely different dicts.

If we think that needs to be revisited, I might want to do that as a followon PR.

src/site_upload/process_flat/process_flat.py
dogversioning: See above - `meta_type` redirection.

src/site_upload/process_upload/process_upload.py
dogversioning: yeah, that's really good idea. #143

src/shared/awswrangler_functions.py
mikix: Maybe you do? I just said that based on how Vlad keeps talking about quotes casually in convos, but maybe in some spots this is the right quoting.

src/site_upload/process_flat/process_flat.py
mikix: OK got it... Maybe this pattern would work better for my brain if the "where does this data get written" field had more prominence - like maybe as a mandatory kwarg or a positional arg, so it's easier to compare two calls and see "ah yes, this arg is different". In the calls above, the first positional arg felt like the "destination" info and the kwargs the values.

Honestly, a lot of these kwargs could use more clarity to a newcomer. Like... "write_local_metadata(value=x, metadata=y)" -- is `value=` the metadata being written and `metadata=` is metadata about the metadata? Or is `metadata=` the metadata being written and `value=` is something else.

dogversioning: yeah - part of my struggle in getting onboarded back to this was remembering what all this stuff does.

This is sort of address via #70, but the names of all these things could use a second pass.

PR #144, opened 2024-12-17
Added study distribution

dogversioning:
This PR adds a mechanism for shipping prepared studies from github directly to sites, by adding a new lambda and an SNS queue with cross-account subscription permissions.


src/post_distribute.Dockerfile
dogversioning, line 1: This dockerfile is where it is so it can correctly access the contents of the `shared` folder. I'm hoping this will be the only one we need to use in this way.

src/post_distribute.Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+WORKDIR ${LAMBDA_TASK_ROOT}
+COPY dashboard/post_distribute/requirements.txt .
+RUN pip install -r requirements.txt
+COPY dashboard/post_distribute/post_distribute.py .
+COPY shared shared
+
+# Force setup of some initial matplotlib configuration artifacts
+RUN mkdir /tmp/matlplotlib
+ENV MPLCONFIGDIR=/tmp/matlplotlib
+RUN cumulus-library version
```
dogversioning, line 12: This was stupid enough that it almost caused me to pull matplotlib from the library and move the PSM charting to a different library. I may still do this.

Otherwise - matplotlib freaks out over not having home access and tries to find a new place to put things, and this often blows our 10 second window for initializing the lambda.

template.yaml
```diff
@@ -742,6 +744,53 @@ Resources:
       LogGroupName: !Sub "/aws/lambda/${DashboardGetStudyPeriodsFunction}"
       RetentionInDays: !Ref RetentionTime
 
+  DashboardPostDistributeStudyFunction:
+    Type: AWS::Serverless::Function
+    Properties:
+      FunctionName: !Sub 'CumulusAggDashboardDistributeStudy-${DeployStage}'
+      PackageType: Image
+      ImageConfig:
+        Command: ["post_distribute.distribute_handler"]
```
dogversioning, line 753: This (and the contents of the meta fields below) are the bits that switch the lambda from the normal execution environment to a container-based execution.

template.yaml
```diff
@@ -787,6 +836,30 @@ Resources:
         - Key: Name
           Value: !Sub 'SNSTopicCacheAPI-${DeployStage}'
 
+  SNSStudyPayload:
+    Type: AWS::SNS::Topic
+    Properties:
+      ContentBasedDeduplication: True
+      FifoTopic: True
```
dogversioning, line 843: This addition does two things:
- insures that a message won't be delivered twice
- insures delivery order to all consumers

mikix (approved):

pyproject.toml
```diff
@@ -46,7 +47,8 @@ test = [
     "moto[s3,athena,sns] == 4.1.5",
     "pytest",
     "pytest-cov",
-    "pytest-mock"
+    "pytest-mock",
+    "responses"
```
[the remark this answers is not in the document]
mikix: I'm still a little mad that JSON won the interchange format wars. It's so annoying and has trained us to think we don't **deserve** niceties like trailing commas and comments.

src/dashboard/post_distribute/post_distribute.py
```diff
@@ -0,0 +1,89 @@
...
+
+# lambda performance tuning - moving these outside mean that they will not
+# contribute to the lambda init window
+sns_client = boto3.client("sns", os.environ.get("AWS_REGION"))
+# in dockerized lambdas, `/tmp` is the only valid write location
+BASE_DIR = "/tmp"  # noqa: S108
+
+
+def get_study_from_github(url):
+    if "smart-on-fhir" not in url:
```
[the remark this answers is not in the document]
mikix: Getting most recent tag could work. I was just nervous around implications for you and me of tip-grabbing. Might cause us to invert our normal dev model and have like, a dev branch that we occasionally copied to `main` when it was stable.

This handler could also accept/require a git revision argument (in addition to a URL), that would be used when cloning. Could be a branch name like `main`, a tag, or a commit ID...

src/dashboard/post_distribute/post_distribute.py
```diff
@@ -0,0 +1,89 @@
+import json
```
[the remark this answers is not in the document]
dogversioning: ok, added some documentation (now in queue_distribute).

src/dashboard/post_distribute/post_distribute.py
```diff
@@ -0,0 +1,89 @@
...
+import os
+import pathlib
+
+import boto3
+import requests
+from cumulus_library import cli
```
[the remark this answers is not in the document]
dogversioning: ok, cut over to subprocess complete (over in queue_distribute)

src/dashboard/post_distribute/post_distribute.py
```diff
@@ -0,0 +1,89 @@
...
+from shared import decorators, functions
+
+# lambda performance tuning - moving these outside mean that they will not
+# contribute to the lambda init window
```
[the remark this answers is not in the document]
dogversioning: tried to clean these up (now in queue_distribute)

src/dashboard/post_distribute/post_distribute.py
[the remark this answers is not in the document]
dogversioning: ok, now checks for a string starting with a github URL and for non-url valid characters.

src/dashboard/post_distribute/post_distribute.py
```diff
@@ -0,0 +1,89 @@
...
+    )
+    raw_url_base = (
+        url.replace("https://github.com/", "https://raw.githubusercontent.com/") + "main/"
+    )
+    study_name = url.split("/")[-2]
+    res = requests.get(api_url, timeout=10)
+    if res.status_code != 200:
+        raise ValueError(f"{url} is not a valid git repository")
+    files = res.json()["tree"]
+    for file in files:
```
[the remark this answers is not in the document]
dogversioning: now just using git

dogversioning (commented):
Addressed some PR comments.

Note that this also decouples 'i receieved a message' from 'i'm taking a long while to build a study', since API gateway automatically throws a timeout error at 29 seconds of runtime.

src/dashboard/post_distribute/post_distribute.py
```diff
@@ -0,0 +1,51 @@
...
+import boto3
+import requests
+
+from shared import functions
+
+sns_client = boto3.client("sns", os.environ.get("AWS_REGION"))
+valid_chars = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789-._~:/?#[]@!$&'()*+,"
+
+
+def validate_github_url(config):
```
mikix, line 15: nit: minor naming bump - this not only validates but also does the request - which might be surprising if you just saw the function name.

src/dashboard/post_distribute/post_distribute.py
```diff
@@ -0,0 +1,51 @@
...
+    if not config["url"].startswith("https://github.com/smart-on-fhir/") or any(
+        c not in valid_chars for c in config["url"]
+    ):
+        raise ValueError(f"{config['url']} is not an official Cumulus study.")
+    res = requests.get(config["url"], timeout=10)
+    if res.status_code != 200:
+        raise ValueError(f"{config['url']} is not a valid git repository")
+    if "tag" in config and config["tag"] is not None:
```
mikix, line 23: If tag is not provided, this method just silently does nothing - maybe throw a ValueError or similar?

Also - do you really want `is None` here and allow an empty string tag? Maybe just `if config.get("tag");` is sufficient.

src/dashboard/queue_distribute/queue_distribute.py
```diff
@@ -0,0 +1,102 @@
...
+# https://docs.aws.amazon.com/lambda/latest/dg/best-practices.html#function-code
+
+sns_client = boto3.client("sns", os.environ.get("AWS_REGION"))
+BASE_DIR = "/tmp"  # noqa: S108
+
+
+def get_study_from_github(config):
+    try:
+        subprocess.run(["/usr/bin/git", "clone", config["url"], f"{BASE_DIR}/studies"], check=True)  # noqa: S603
+        if "tag" in config and config["tag"] is not None:
```
mikix, line 43: similar "if no tag, just silently does nothing" behavior here.

src/dashboard/queue_distribute/queue_distribute.py
```diff
@@ -0,0 +1,102 @@
...
+            subprocess.run(  # noqa: S603
+                ["/usr/bin/git", "checkout", "tag"],
```
mikix, line 45: I [think](https://stackoverflow.com/questions/20280726/how-to-clone-a-specific-git-tag) you can pass `--branch <tag>` to the clone command instead of doing a separate checkout (and maybe also `--depth 1` for performance).

src/dashboard/post_distribute/post_distribute.py
```diff
@@ -0,0 +1,51 @@
...
+from shared import decorators, functions
+
+sns_client = boto3.client("sns", os.environ.get("AWS_REGION"))
+valid_chars = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789-._~:/?#[]@!$&'()*+,"
+
+
+def validate_github_url(config):
+    if not config["url"].startswith("https://github.com/smart-on-fhir/") or any(
+        c not in valid_chars for c in config["url"]
+    ):
```
mikix, line 18: I _think_ this is safe, but I'm not like 100% confident that there isn't some valid URL syntax that lets you have `github.com/smart-on-fhir/` as your username and then the real domain afterward :smile: 

I'd feel lightly safer if you used urllib.parse and then checked netloc and path. But not critical. I couldn't break this with some light testing.

mikix (approved):

src/dashboard/post_distribute/post_distribute.py
dogversioning: to my mind, 'validate' in this context is 'validate the url is actually a git repository'. Open to tweaks.

src/dashboard/post_distribute/post_distribute.py
dogversioning: this was intentional - if you don't provide a tag, you want the tip of main. can be less kind if you think it's error prone.

src/dashboard/queue_distribute/queue_distribute.py
dogversioning: ok, perf, i'll take it

src/dashboard/post_distribute/post_distribute.py
dogversioning: if you want - i went looking into urllib.parse's behavior and it specifically does not do the character kind of validation, so I think this is stricter than that? but can't hurt.

src/dashboard/post_distribute/post_distribute.py
mikix: Oh I guess this was doing more than just pinging URLs and actually doing some downloads. I think I just misread it.

src/dashboard/post_distribute/post_distribute.py
mikix: Oh I guess I thought this was doing more than just pinging the URLs and actually doing downloads. I think I  just misread it.

src/dashboard/post_distribute/post_distribute.py
mikix: Yeah that's fine, it goes back to my confusion earlier - where like this get request _did_ something. I see now what's up. :+1:

PR #145, opened 2025-01-31
Improved handling for filters

dogversioning:
updates handling of a case when a filter parameter is not in a displayed column in the dashboard, removing it from the set of columns we use for filtering out non-nulls.

### DIFF
--- a/src/dashboard/queue_distribute/queue_distribute.py
+++ b/src/dashboard/queue_distribute/queue_distribute.py
@@ -40,8 +40,8 @@ BASE_DIR = "/tmp"  # noqa: S108
 def get_study_from_github(config):
     try:
         args = ["--depth", "1", config["url"], f"{BASE_DIR}/studies"]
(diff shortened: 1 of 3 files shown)


src/dashboard/queue_distribute/queue_distribute.py
```diff
@@ -40,7 +40,7 @@
 def get_study_from_github(config):
     try:
         args = ["--depth", "1", config["url"], f"{BASE_DIR}/studies"]
-        if config["tag"]:
+        if config.get("tag"):
```
dogversioning, line 43: this is an unrelated change i had hanging around locally from the last time i was working on message queues.

mikix: Looks like a prime spot for a walrus operator...  :eyes: :smile:

mikix (approved):

PR #148, opened 2025-02-20
Test chart data endpoint w/ duckdb

dogversioning:
This adds a duckdb fixture to the aggregator and uses it for testing the chart data endpoint, and fixes several issues found as a result.


src/dashboard/get_chart_data/get_chart_data.py
```diff
@@ -196,6 +198,7 @@ def _build_query(query_params: dict, filter_groups: list, path_params: dict) ->
             coalesce_columns=columns,
             inline_configs=inline_configs,
             none_configs=none_configs,
+            extra_filter_configs=[],  # extra_filter_configs
```
mikix, line 201: What's this line about? Did you mean to leave that comment like that?

src/site_upload/cache_api/cache_api.py
```diff
@@ -38,13 +38,19 @@ def cache_api_data(s3_client, s3_bucket_name: str, db: str, target: str) -> None
...
         for version in versions:
+            if version != dp:
+                continue
             dp_dict = {
                 **dp_detail,
                 **versions[version],
-                "version": version,
-                "id": f"{dp_detail['study']}__{dp_detail['name']}__{version}",
+                "version": version.split("__")[2],
+                "id": f"{dp_detail['study']}__{dp_detail['name']}__{version.split('__')[2]}",
```
mikix, line 53: nit: Is `version` worth splitting separately into a variable, for clarity?

Like `_version_package, version_number = version.split("__", 2)` or something. (for example, I don't know what the first part of the version is, so I just made up a variable name)

tests/conftest.py
```diff
@@ -176,6 +179,36 @@ def mock_db():
...
+
+    def _compat_from_iso8601_timestamp(
+        value: str | datetime.datetime,
+    ) -> datetime.datetime:
+        if type(value) is str:
+            return datetime.datetime.fromisoformat(value)
+        return value
+
+    db = duckdb.connect(tmp_path / "duck.db")
+    db.create_function(
```
mikix, line 196: Obviously you thought about this, but let's spill a tiny amount of digital ink. Pros/cons of having tests depend on cumulus-library and grabbing its DuckdbDatabase class? Not worth it yet, I'm guessing?

mikix (approved):

src/dashboard/get_chart_data/get_chart_data.py
dogversioning: That was from an in process attempt to fix the underlying ginger template, which moved away from; will delete

src/site_upload/cache_api/cache_api.py
dogversioning: Yeah, I buy it

tests/conftest.py
dogversioning: Yeah, a couple thoughts

- the aggregator doesn't have Python version matrix compatibility problems. It runs on a specific version and there's some workaround code for Python 311 in there that just doesn't apply here.
  - there's one weird case where we want this to perform slightly differently around, just not caring about finding real time stamps in the database with test data
- we're only dealing with this one function for now, so the library class is a little heavy and having its own thing provides a nice isolation. If that changed that I might feel more strongly about importing that class.

Issue #143, opened 2024-12-04
Centralize path formatting in a single class


dogversioning:
We're doing a ton of path slicing, in a lot of different places. We should have a single class that manages all the 'slice these s3 urls in this way' logic.

dogversioning:
addressed via #158

PR #149, opened 2025-03-07
Generate formats from parquet

dogversioning:
This PR makes the following changes:

- Added a new endpoint, /from-parquet/, which takes an `s3_path` path parameter (present in the data packages endpoint) to return data transformed from that format
  - This supports an optional `type` path parameter - if not provided, json is returned (which is the expected use case for now). `csv` and `tsv` can also be provided for those export formats (as can `json`)
  - the parquet path is now returned when constructing the data packages endpoint instead of the csv endpoint
  - the code related to getting last valid/aggregate csvs previously were unused, and have been removed
- flat files update metadata on each load, rather than first time only
- some documentation on how to run in watch mode updated to behave better with the addition of docker-based lambdas.


docs/dashboard_api.prod.yaml
```diff
@@ -551,6 +396,27 @@ paths:
           content: {}
       security:
       - api_key: []
+  /from-parquet:
```
mikix, line 399: Your commit message has from_parquet, so I was happy to see the hyphen here 😄

src/dashboard/get_from_parquet/get_from_parquet.py
```diff
@@ -0,0 +1,33 @@
...
+    """manages event and returns parquet file as json dict"""
+    del context
+    s3_bucket_name = os.environ.get("BUCKET_NAME")
+    try:
+        s3_path = event["queryStringParameters"]["s3_path"]
+        df = awswrangler.s3.read_parquet(f"s3://{s3_bucket_name}/{s3_path}")
+    except (KeyError, FileNotFoundError, pyarrow.lib.ArrowInvalid):
+        res = functions.http_response(404, "S3_path not found")
+        return res
+    if output_type := event["queryStringParameters"].get("type"):
```
mikix, line 20: This path doesn’t allow for passing type=json

src/site_upload/cache_api/cache_api.py
```diff
@@ -43,17 +44,23 @@ def cache_api_data(s3_client, s3_bucket_name: str, db: str, target: str) -> None
...
+                version = dp_id
             dp_dict = {
                 **dp_detail,
                 **dp_ids[dp_id],
                 "version": version,
                 "id": f"{dp_detail['study']}__{dp_detail['name']}__{version}",
             }
-            if "__flat" in dp:
+            if "__flat" in dp_dict["s3_path"]:
+                site = dp_dict["s3_path"].split("/")[5]
```
mikix, line 60: More magic path numbering 😕

dogversioning: there's an issue about it - #143 - I just don't have the time right now with the backlog of other things i've got

src/dashboard/get_from_parquet/get_from_parquet.py
dogversioning: i went back and forth on this - i'll add an explicit case

Issue #150, opened 2025-03-07
Data transmission timestamp management


James-R-Jones:
Expected behavior: dashboard reflects most recent data transmission timestamps for data sent to the aggregator

Current behavior: timestamp metadata may only be updating the first time a dataset is sent per study version. 

Could future proof this a bit by adding support to track historical transmission logs with a bit more metadata - (site, study, version, timestamp, {total count per aggregate slice}). Probably better in a separate feature request.

PR #153, opened 2025-03-11
Update study period, add total to flatfile summary

dogversioning:
- Adresses cases where study_period would not update (#150)
- Adds `total` to flat file summary in data packages
- Updates evergreen migration to support flat files


template.yaml
```diff
@@ -1039,24 +1039,24 @@ Resources:
...
+    #      AssumeRolePolicyDocument:
+    #        Version: '2012-10-17'
+    #        Statement:
+    #          Action: 'sts:AssumeRole'
+    #          Effect: Allow
+    #          Principal:
+    #            Service: apigateway.amazonaws.com
+    #      Path: /
+    #      ManagedPolicyArns:
+    #        - 'arn:aws:iam::aws:policy/service-role/AmazonAPIGatewayPushToCloudWatchLogs'
```
dogversioning, line 1059: dangling commenting from the last PR

scripts/migrations/migration.003.data_packages.py
dogversioning, line 1: as this gets pulled along - is it worth considering a different class of maintenance object than migrations? i could see this being part of a group of bucket management tools, which could import thing directly rather than duplicating them directly here.

src/shared/awswrangler_functions.py
```diff
@@ -14,8 +14,15 @@ def get_s3_data_package_list(
...
     extension: str = "parquet",
+    version: str | None = None,
+    site: str | None = None,
 ):
     """Retrieves a list of data packages for a given S3 path post-upload processing"""
+    if bucket_root in [BucketPath.FLAT.value, BucketPath.CSVFLAT.value]:
+        return awswrangler.s3.list_objects(
+            path=f"s3://{s3_bucket_name}/{bucket_root}/{study}/{site}/",
+            suffix=extension,
+        )
```
dogversioning, line 25: not used anyplace currently, was just futureproofing while i was in here.

src/shared/functions.py
```diff
@@ -225,7 +225,12 @@ def get_s3_site_filename_suffix(s3_path: str):
     # The expected s3 path for site data packages looks like:
     #   s3://bucket_name/enum_value/site/study/data_package/file
     # so this is returning data_package/file
-    return "/".join(s3_path.split("/")[6:])
+    return "/".join(s3_path.split("/")[-3:])
```
dogversioning, line 228: this is also a future compatibility change

src/shared/functions.py
```diff
@@ -225,7 +225,12 @@ def get_s3_site_filename_suffix(s3_path: str):
...
+
+
+def get_s3_key_from_path(s3_path: str):
+    """returns a valid S3 key given an S3 path"""
+    return "/".join(s3_path.split("/")[3:])
```
dogversioning, line 233: this is slowly slicing off bits of #143

src/shared/pandas_functions.py
```diff
@@ -13,23 +13,25 @@ def get_column_datatypes(dtypes: pandas.DataFrame):
...
+            "uint64",
         ):
             column_dict[column] = "integer"
-        elif str(dtypes[column]) in ("Float32", "Float64"):
+        elif str(dtypes[column]).lower() in ("float32", "float64"):
             column_dict[column] = "float"
         elif str(dtypes[column]) == "boolean":
             column_dict[column] = "boolean"
+        elif column in ["median", "average", "std_dev", "percentage"]:
+            column_dict[column] = "double"
```
mikix, line 34: I know we talked about this - and this is fine for now. I'm just curious - what is the dtype coming across as for these decimal columns (at least, decimal when written by data-metrics originally)?

src/site_upload/study_period/study_period.py
```diff
@@ -11,12 +11,23 @@
...
-    path = awswrangler_functions.get_s3_study_meta_list(
+    paths = awswrangler_functions.get_s3_study_meta_list(
         s3_bucket, study, data_package, site, version
     )
-    if len(path) != 1:
-        raise KeyError("Unique date path not found")
-    df = awswrangler.s3.read_parquet(path[0])
+    if len(paths) > 1:
+        latest_key = None
+        latest_date = datetime(1000, 1, 1, tzinfo=UTC)
```
mikix, line 19: There is `datetime.min` (`== datetime.datetime(1, 1, 1, 0, 0)`) which might the intent here more obvious.

src/shared/pandas_functions.py
dogversioning: object, frustratingly.

In parquet, the type is showing up as FIXED_LEN_BYTE_ARRAY, with a note that DECIMAL is the legacy type - might be something that would be addressed by a pandas update?

src/site_upload/study_period/study_period.py
```diff
@@ -11,12 +11,23 @@
...
+        for path in paths:
+            key = functions.get_s3_key_from_path(path)
+            res = s3_client.head_object(Bucket=s3_bucket, Key=key)
+            if res["LastModified"] > latest_date:
+                latest_key = key
+                latest_date = res["LastModified"]
```
mikix, line 25: nit: or here's a way to avoid tracking latest_xxx variables:
```
def modified_time(path):
    key = functions.get_s3_key_from_path(path)
    res = s3_client.head_object(Bucket=s3_bucket, Key=key)
    return res["LastModified"]

latest_path = max(paths, key: modified_time)
```

src/site_upload/study_period/study_period.py
```diff
@@ -11,12 +11,23 @@
...
+        for path in paths:
+            if latest_key not in path:
```
mikix, line 27: I always get so nervous when I saw a bare `in` check like this - worried about false positives. Would you expect multiple paths to match normally? Can we use `latest_path` instead of `latest_key` and then just do an equality check? Or if we like keeping the key only, maybe use `endswith` instead of `in`?

mikix (approved):

src/site_upload/study_period/study_period.py
dogversioning: so since these are file paths inside of a bucket, and we're not having users upload directly, they should be globally unique and unlikely to be substrings

i can use path instead, but key is the thing that the boto tools expect, so that's what we want to use later.

scripts/migrations/migration.003.data_packages.py
dogversioning: this one, specifically, was meant to be a one and done, and then i kept finding it solved a ton of problems. And to your point - it's not really a migration then, so promotion makes sense. Replacing with a queen.

src/site_upload/study_period/study_period.py
mikix: Sure, but you don't use `latest_key` later in this case. But I get ya.

dogversioning: yeah, i ended up pulling it out anyway as part of moving to max(function) as you suggested

PR #156, opened 2025-03-17
Promote some migrations to maintenance scripts

dogversioning:
This makes the following changes:
- Moves reused migrations to two more permanent maintenance scripts
  - Reuses code from aggregator rather than creating parallel infrastructure, where possible
  - Changes the packager to better support these imports
- Fixes some sloppy naming/copy errors in migration 6, for posterity
- Gets correct value for site names in caching flat files
- Updates some csv headers by request

### DIFF
--- a/scripts/delete_site_data.py
+++ b/scripts/delete_site_data.py
@@ -1,0 +1,126 @@
+import argparse
+from collections import defaultdict
+
+import boto3
+from rich import console, progress, table
+
+from src.shared import enums
+
+site_artifacts = [enums.BucketPath.LAST_VALID.value]
+aggregates = [enums.BucketPath.AGGREGATE, enums.BucketPath.CSVAGGREGATE.value]
+
+
+def get_subbucket_contents(client, bucket, prefix):
+    res = client.list_objects_v2(Bucket=bucket, Prefix=prefix)
+    if "Contents" not in res.keys():
+        return []
+    files = [x["Key"] for x in res["Contents"]]
+    while res["IsTruncated"]:
+        res = client.list_objects_v2(
+            Bucket=bucket, Prefix=prefix, ContinuationToken=res["NextContinuationToken"]
+        )
+        files += [x["Key"] for x in res["Contents"]]
+    return files
+
+
+def cleanup_target(tree: dict, target: str, data_packages: str, site: str, version: str):
+    other_sites = list(tree[data_packages].keys())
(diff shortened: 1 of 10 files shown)


src/dashboard/get_from_parquet/get_from_parquet.py
```diff
@@ -30,7 +30,16 @@ def from_parquet_handler(event, context):
...
         return functions.http_response(
-            200, payload, extra_headers={"Content-Type": "text/csv"}, skip_convert=True
+            200,
+            payload,
+            extra_headers={
+                "Content-Type": "text/csv",
+                "Content-disposition": (
+                    f"attachment; filename={s3_path.split('/')[-1].replace('.parquet','.csv')}"
+                ),
+                "Content-Length": len(payload.encode("UTF-8")),
```
mikix, line 40: This feels odd - I would expect `len(payload)` (i.e. the length of the bytes sent, not the resulting utf8 string)

src/site_upload/cache_api/cache_api.py
```diff
@@ -57,7 +57,7 @@ def cache_api_data(s3_client, s3_bucket_name: str, db: str, target: str) -> None
                 "id": f"{dp_detail['study']}__{dp_detail['name']}__{version}",
             }
             if "__flat" in dp_dict["s3_path"]:  # pragma: no cover
-                site = dp_dict["s3_path"].split("/")[5]
+                site = dp_dict["s3_path"].split("/")[2]
```
mikix, line 60: image

scripts/delete_site_data.py
```diff
@@ -0,0 +1,126 @@
+import argparse
```
mikix, line 1: This is a copy of migration 5? Why not do a git-rename / delete of that file? (same question for the other script)

mikix (approved):

src/site_upload/cache_api/cache_api.py
dogversioning: it's on my list, i know you hate it

scripts/delete_site_data.py
mikix: Maybe a comment at the top of those files then? If a rando like me comes by in a year, I'll see two very similar files and be unclear on the history/reason. Though I suppose I can go to git log for that.

dogversioning: ok, done.

PR #158, opened 2025-03-20
Added s3 key parser

dogversioning:
This moves all the 'get site/study/etc from a key' to a shared function, and changes site upload modules to use it.

It also adds a mock flat file metadata entry to the bucket mock.

### DIFF
--- a/src/shared/functions.py
+++ b/src/shared/functions.py
@@ -275,10 +275,9 @@ class PackageMetadata:
     version: str
 
 
-def parse_s3_key(key: str) -> dict:
+def parse_s3_key(key: str) -> PackageMetadata:
     """Handles extraction of package metadata from an s3 key"""
     key = key.split("/")
-    print(key)
     match key[0]:
         case enums.BucketPath.AGGREGATE.value | enums.BucketPath.CSVAGGREGATE.value:
             package = PackageMetadata(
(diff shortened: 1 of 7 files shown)


src/shared/functions.py
```diff
@@ -264,3 +265,56 @@ def get_latest_data_package_version(bucket, prefix):
...
+    study: str
+    site: str | None
+    data_package: str
+    version: str
+
+
+def parse_s3_key(key: str) -> dict:
+    """Handles extraction of package metadata from an s3 key"""
+    key = key.split("/")
+    print(key)
```
mikix, line 281: :scissors:

src/shared/functions.py
```diff
@@ -264,3 +265,56 @@ def get_latest_data_package_version(bucket, prefix):
...
+
+@dataclasses.dataclass()
+class PackageMetadata:
+    study: str
+    site: str | None
+    data_package: str
+    version: str
+
+
+def parse_s3_key(key: str) -> dict:
```
mikix, line 278: This doesn't return a dict, but `PackageMetadata`

src/shared/functions.py
```diff
@@ -264,3 +265,56 @@ def get_latest_data_package_version(bucket, prefix):
...
+            | enums.BucketPath.FLAT.value
+            | enums.BucketPath.LATEST.value
+        ):
+            package = PackageMetadata(
+                study=key[1],
+                site=key[3],
+                data_package=key[2].split("__")[1],
+                version=key[4],
+            )
+        case enums.BucketPath.ERROR.value | enums.BucketPath.LAST_VALID.value:
```
mikix, line 302: This branch looks identical to the above branch. Any reason to duplicate?

mikix (approved):
The other cases of `split("/")` in the codebase don't need converting to this?

src/shared/functions.py
dogversioning: so it is! it didn't start out that way, but will consolidate.

Issue #166, opened 2025-05-06
Bug: chart_data endpoint getting a numpy int type when column is an integer


dogversioning:
This is messing with dictionary retrieval. we should try and detect this case and cast the value appropriately.

traceback:

```traceback: [{'filename': '/var/task/shared/decorators.py', 'name': 'wrapper', 'lineno': 16}, {'filename': '/var/task/get_chart_data.py', 'name': 'chart_data_handler', 'lineno': 262}, {'filename': '/var/task/shared/functions.py', 'name': 'http_response', 'lineno': 62}, {'filename': '/var/lang/lib/python3.11/json/__init__.py', 'name': 'dumps', 'lineno': 238}, {'filename': '/var/lang/lib/python3.11/json/encoder.py', 'name': 'encode', 'lineno': 200}, {'filename': '/var/lang/lib/python3.11/json/encoder.py', 'name': 'iterencode', 'lineno': 258}]",```

Issue #165, opened 2025-05-06
Bug: chart data timing out for large datasets


dogversioning:
Depending on the filter configuration, the get_chart_data endpoint may have a lot of data in memory - this can cause dataframe actions to run for a minute or more.

This :might: be fixed by adding more memory to the lambda, or investigating how the data frame is otherwise being constructed.

dogversioning:
Noting that the `_format_payload` function may be a source of memory intensive activities as well.

PR #167, opened 2025-05-07
Chart data endpoint bugfixes

dogversioning:
- Replaces very slow slicing & masking with builtin groupby function on dataframes, which improves execution time by 100x for large datasets(!!) #165 
- Casts index values so that the json serializer doesn't choke on numpy integers/bools #166

### DIFF
--- a/tests/test_data/cube_response_filtered_stratified.json
+++ b/tests/test_data/cube_response_filtered_stratified.json
@@ -14,8 +14,8 @@
             "stratifier": "",
             "rows": [
                 [
-                    16917,
-                    "female"
+                    "female",
+                    16917
                 ]
             ]
         },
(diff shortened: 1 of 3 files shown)


tests/test_data/cube_response_filtered_stratified.json
```diff
@@ -14,44 +14,44 @@
             "stratifier": "",
             "rows": [
                 [
-                    16917,
-                    "female"
```
dogversioning, line 18: this was always supposed to be a label/number ordered pairing - this ordering was a product of the old assembly method and our test data. so the new method is handling an unexpected potential bug case (great!). but that's the reason for the in-array ordering change.

The actual order of the rows themselves doesn't matter, so not worrying too much about the shift in order of the individual rows in some cases.

PR #168, opened 2025-05-08
Back to strings for column versions

dogversioning:
After seeing this pop up again, i did some global searches on versions - I think the change of string to int (which was only being done in the migrations) is more disruptive than changing int to string, so every instance now is configured to string versions.

### DIFF
--- a/scripts/migrations/migration.003.data_packages.py
+++ b/scripts/migrations/migration.003.data_packages.py
@@ -104,7 +104,7 @@ def update_column_type_metadata(bucket: str):
             output.setdefault(study, {})
             output[study].setdefault(data_package, {})
             output[study][data_package].setdefault(version, {})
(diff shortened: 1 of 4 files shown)


scripts/migrations/migration.003.data_packages.py
```diff
@@ -104,7 +104,7 @@ def update_column_type_metadata(bucket: str):
             output.setdefault(study, {})
             output[study].setdefault(data_package, {})
             output[study][data_package].setdefault(version, {})
-            output[study][data_package][version]["column_types_format_version"] = 2
+            output[study][data_package][version]["column_types_format_version"] = "2"
```
dogversioning, line 107: i realize this is historical but i'm just trying to squash all of these and be done with it.

mikix (approved):

PR #171, opened 2025-05-12
Handle multiple files in previous uploads, remove CSVs

dogversioning:
Broken up into three commits for my own sanity/testing isolation:
Commit one: 
 - Removes CSV generation, and updates infrastructure to not care about csv paths/not expect csvs during tests
 - Adds a migration for cleaning up buckets
Commit two:
 - Changes how archiving works during powerset merge to archive everything in a folder, rather than expecting a filename pattern
 Commit three:
 - Removes the CSV enums, and also some ruff formatting on push (probably fine to skim for review

### DIFF
--- a/scripts/migrations/migration.008.remove_csvs.py
+++ b/scripts/migrations/migration.008.remove_csvs.py
@@ -1,0 +1,41 @@
+import argparse
+import enum
+
+import boto3
+from rich import progress
+
+
+class BucketPath(enum.Enum):
+    """stores root level buckets for managing data processing state"""
+
+    ADMIN = "admin"
+    AGGREGATE = "aggregates"
+    ARCHIVE = "archive"
+    CACHE = "cache"
+    CSVAGGREGATE = "csv_aggregates"
+    CSVFLAT = "csv_flat"
+    ERROR = "error"
+    FLAT = "flat"
+    LAST_VALID = "last_valid"
+    LATEST = "latest"
+    META = "metadata"
+    STUDY_META = "study_metadata"
+    UPLOAD = "site_upload"
+
+
+def remove_csvs(bucket):
+    client = boto3.client("s3")
+    # We specifically want to look at just the BucketPath folders since much of the
+    # raw athena data is stored in csvs, and we don't want to touch that
+    for base_folder in progress.track(BucketPath, description="Cleaning csvs..."):
+        res = client.list_objects_v2(Bucket=bucket, Prefix=base_folder.value)
+        for file in res.get("Contents", []):
+            if file["Key"].endswith(".csv"):
+                client.delete_object(Bucket=bucket, Key=file["Key"])
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description="""Removes aggregator generated csvs. """)
+    parser.add_argument("-b", "--bucket", help="bucket name")
+    args = parser.parse_args()
+    remove_csvs(args.bucket)
--- a/src/shared/functions.py
+++ b/src/shared/functions.py
@@ -224,7 +224,7 @@ def get_s3_keys(
     return contents
 
 
-def get_s3_filename(s3_path: str):
+def get_filename_from_s3_path(s3_path: str):
     """Given an s3 path/key, returns the filename"""
     return s3_path.split("/")[-1]
 
--- a/src/shared/functions.py
+++ b/src/shared/functions.py
@@ -224,12 +224,9 @@ def get_s3_keys(
     return contents
 
 
-def get_s3_site_filename_suffix(s3_path: str):
-    """Extracts site/filename data from s3 path"""
-    # The expected s3 path for site data packages looks like:
-    #   s3://bucket_name/enum_value/site/study/data_package/file
-    # so this is returning data_package/file
-    return "/".join(s3_path.split("/")[-3:])
+def get_s3_filename(s3_path: str):
+    """ Given an s3 path/key, returns the filename"""    
+    return s3_path.split('/')[-1]
 
 
 def get_s3_key_from_path(s3_path: str):
@@ -283,6 +280,8 @@ class PackageMetadata:
 def parse_s3_key(key: str) -> PackageMetadata:
     """Handles extraction of package metadata from an s3 key"""
     try:
+        #did we get a full path instead?
+        key = get_s3_key_from_path(key)
         key = key.split("/")
         match key[0]:
             case enums.BucketPath.AGGREGATE.value | enums.BucketPath.CSVAGGREGATE.value:
--- a/src/shared/functions.py
+++ b/src/shared/functions.py
@@ -225,8 +225,8 @@ def get_s3_keys(
 
 
 def get_s3_filename(s3_path: str):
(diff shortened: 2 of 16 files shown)


src/shared/functions.py
```diff
@@ -224,12 +224,9 @@ def get_s3_keys(
...
 
 
-def get_s3_site_filename_suffix(s3_path: str):
-    """Extracts site/filename data from s3 path"""
-    # The expected s3 path for site data packages looks like:
-    #   s3://bucket_name/enum_value/site/study/data_package/file
-    # so this is returning data_package/file
-    return "/".join(s3_path.split("/")[-3:])
+def get_s3_filename(s3_path: str):
+    """ Given an s3 path/key, returns the filename"""
```
mikix: nit: being slightly pedantic, you could probably just call this `get_filename` (since there's nothing S3 about it) - but this is also fine.

src/site_upload/powerset_merge/powerset_merge.py
```diff
@@ -118,56 +118,67 @@ def merge_powersets(manager: s3_manager.S3Manager) -> None:
...
-            if any(x.endswith(site_specific_name) for x in last_valid_file_list):
-                manager.copy_file(
-                    f"{enums.BucketPath.LAST_VALID.value}/{subbucket_path}",
-                    f"{enums.BucketPath.ARCHIVE.value}/{timestamped_path}",
-                )
+            if matches := filter(lambda x: subbucket_path in x, last_valid_file_list):
+                date_str = datetime.datetime.now(datetime.UTC).isoformat()
+                for match in matches:
+                    match_filename = functions.get_s3_filename(match)
+                    match_timestamped_filename = f".{date_str}.".join(match_filename.split("."))
```
mikix, line 135: nit: weird future-behavior bug here if we ever use a filename with two dots in it.

src/site_upload/powerset_merge/powerset_merge.py
```diff
@@ -118,56 +118,67 @@ def merge_powersets(manager: s3_manager.S3Manager) -> None:
...
+                    last_target = (
+                        f"{enums.BucketPath.LAST_VALID.value}/{subbucket_path}/{match_filename}"
+                    )
```
mikix, line 138: nit: could this just be `last_target = match`? Just to avoid a mistake when reconstructing the path.

src/site_upload/powerset_merge/powerset_merge.py
```diff
@@ -118,56 +118,67 @@ def merge_powersets(manager: s3_manager.S3Manager) -> None:
...
+                        last_target,
+                        archive_target
+                    )
+                    archives.append((archive_target,last_target))
             # otherwise, this is the first instance - after it's in the database,
             # we'll generate a new list of valid tables for the dashboard
             else:
                 is_new_data_package = True
             df = expand_and_concat_powersets(df, latest_path, manager.site)
+            filename=functions.get_s3_filename(latest_path)
```
mikix: nit: is it worth adding `filename` to the s3 key parser method?

src/site_upload/powerset_merge/powerset_merge.py
```diff
@@ -118,56 +118,67 @@ def merge_powersets(manager: s3_manager.S3Manager) -> None:
...
-            f"{manager.study}/{manager.study}__{manager.data_package}" f"/{site_specific_name}"
-        )
-        date_str = datetime.datetime.now(datetime.UTC).isoformat()
-        timestamped_name = f".{date_str}.".join(site_specific_name.split("."))
-        timestamped_path = (
-            f"{manager.study}/{manager.study}__{manager.data_package}" f"/{timestamped_name}"
+            f"{manager.study}/{manager.study}__{manager.data_package}/{latest_metadata.site}/" 
+            f"{manager.study}__{manager.data_package}__{manager.version}"
         )
+        archives = []
```
mikix, line 127: nit: maybe something that is more descriptive - like `archived_filed` or `moved_files` or something? To indicate that this is a record of what we did, so we can undo it.

src/site_upload/powerset_merge/powerset_merge.py
```diff
@@ -118,56 +118,67 @@ def merge_powersets(manager: s3_manager.S3Manager) -> None:
...
-            if any(x.endswith(site_specific_name) for x in last_valid_file_list):
-                df = expand_and_concat_powersets(
-                    df,
-                    f"s3://{manager.s3_bucket_name}/{enums.BucketPath.LAST_VALID.value}"
-                    f"/{subbucket_path}",
-                    manager.site,
-                )
-                manager.update_local_metadata(enums.TransactionKeys.LAST_AGGREGATION.value)
+            if matches := filter(lambda x: subbucket_path in x, last_valid_file_list):
+                for match in matches:
```
mikix, line 174: nit: similar consolidation here to one line: `for match in filter()`. Same "bare `in` string check" comment too.

mikix (approved):

scripts/migrations/migration.008.remove_csvs.py
```diff
@@ -0,0 +1,41 @@
...
+
+def remove_csvs(bucket):
+    client = boto3.client("s3")
+    # We specifically want to look at just the BucketPath folders since much of the
+    # raw athena data is stored in csvs, and we don't want to touch that
+    for base_folder in progress.track(BucketPath, description="Cleaning csvs..."):
+        res = client.list_objects_v2(Bucket=bucket, Prefix=base_folder.value)
+        for file in res.get("Contents", []):
+            if file["Key"].endswith(".csv"):
+                client.delete_object(Bucket=bucket, Key=file["Key"])
```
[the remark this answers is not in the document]
dogversioning: yeah, the csvs are always generated from parquet, so they have no place else to come from (unless someone is uploading csvs during a study build, but that goes in the athena base dir we're skipping)

src/site_upload/powerset_merge/powerset_merge.py
```diff
@@ -96,16 +94,18 @@ def merge_powersets(manager: s3_manager.S3Manager) -> None:
...
+        subbucket_path = (
+            f"{manager.study}/{manager.data_package}/{last_valid_metadata.site}/"
+            f"{manager.study}__{manager.data_package}__{manager.version}"
+        )
         # If the latest uploads don't include this site, we'll use the last-valid
         # one instead
         try:
-            if not any(x.endswith(site_specific_name) for x in latest_file_list):
-                df = expand_and_concat_powersets(df, last_valid_path, last_valid_site)
+            if not any(subbucket_path in x for x in latest_file_list):
```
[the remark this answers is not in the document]
dogversioning: no, because the latest file will also have the basedir at the front. i could make this a regex but i personally hate that more than this kind of substring check.

src/shared/functions.py
dogversioning: `get_filename_from_s3_path` is probably the more explicit function name

src/site_upload/powerset_merge/powerset_merge.py
dogversioning: yep, sure could

PR #177, opened 2025-05-16
Remove all previous files when a new flat file is uploaded

dogversioning:
In case the format of the flat destination changes, we'll remove all matching files for a given flat data package in a previous upload.

### DIFF
--- a/src/site_upload/process_flat/process_flat.py
+++ b/src/site_upload/process_flat/process_flat.py
@@ -11,15 +11,13 @@ logger.setLevel(log_level)
 
 
 def process_flat(manager: s3_manager.S3Manager):
-    if awswrangler.s3.does_object_exist(
-        f"s3://{manager.s3_bucket_name}/{manager.parquet_flat_key}"
-    ):
-        manager.move_file(
-            manager.parquet_flat_key,
-            manager.parquet_flat_key.replace(
-                enums.BucketPath.FLAT.value, enums.BucketPath.ARCHIVE.value
-            ),
-        )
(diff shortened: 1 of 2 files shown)


src/site_upload/process_flat/process_flat.py
```diff
@@ -11,15 +11,13 @@
...
-    ):
-        manager.move_file(
-            manager.parquet_flat_key,
-            manager.parquet_flat_key.replace(
-                enums.BucketPath.FLAT.value, enums.BucketPath.ARCHIVE.value
-            ),
-        )
+    flat_path = manager.parquet_flat_key[:-1]
+    for key in manager.get_data_package_list(enums.BucketPath.FLAT.value):
+        if flat_path in key:
```
mikix, line 16: hasprefix wouldn't work here?

mikix (approved):

src/site_upload/process_flat/process_flat.py
```diff
@@ -11,15 +11,13 @@
...
-    if awswrangler.s3.does_object_exist(
-        f"s3://{manager.s3_bucket_name}/{manager.parquet_flat_key}"
-    ):
-        manager.move_file(
-            manager.parquet_flat_key,
-            manager.parquet_flat_key.replace(
-                enums.BucketPath.FLAT.value, enums.BucketPath.ARCHIVE.value
-            ),
-        )
+    flat_path = manager.parquet_flat_key[:-1]
```
dogversioning, line 14: anticipating slice complaints, will be better addressed by #178

PR #185, opened 2025-06-25
Stage downloads to S3

dogversioning:
Since lambdas have very tight caps on the size of payloads they can create, we're writing them to S3 and then issuing a pre-signed URL, and letting that handle the download portion.

There's also some light cleanup of some string splicing in shared.functions.

### DIFF
--- a/src/dashboard/get_from_parquet/get_from_parquet.py
+++ b/src/dashboard/get_from_parquet/get_from_parquet.py
@@ -1,7 +1,7 @@
 import awswrangler
 import pyarrow
 
-from shared import decorators, functions
+from shared import decorators, enums, functions, s3_manager
 
 
 @decorators.generic_error_handler(msg="Error retrieving data from parquet")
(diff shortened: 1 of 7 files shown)


src/dashboard/get_from_parquet/get_from_parquet.py
```diff
@@ -17,28 +17,17 @@ def from_parquet_handler(event, context):
...
-            skip_convert=True,
-        )
+                data = df.to_json(orient="table", index=False)
+
     else:
-        return functions.http_response(
-            200, df.to_json(orient="table", index=False), skip_convert=True
-        )
+        data = df.to_json(orient="table", index=False)
+    temp_path = f"{enums.BucketPath.TEMP.value}/{functions.get_s3_key_from_path(s3_path)}"
```
mikix, line 28: Any collision here for multiple requests at same time is fine, because it will be the same file contents, yeah?

src/shared/s3_manager.py
```diff
@@ -28,38 +29,41 @@ class S3Manager:
 
     def __init__(self, event):
         self.s3_bucket_name = os.environ.get("BUCKET_NAME")
-        self.s3_client = boto3.client("s3")
+        self.s3_client = boto3.client("s3", config=boto3.session.Config(signature_version="s3v4"))
```
mikix, line 32: s3v4 is the default already, yeah?

mikix (approved):

PR #190, opened 2025-07-15
Fixed double counting issue in powerset merge

dogversioning:
Addressed the following sources of duplication:
- In cases where a last_valid and latest file were present from the same site, both files were merged, usually doubling the expected value for the most recent upload
- Fixed detection of the last_valid location to match the latest location/expected naming format for uploads

### DIFF
--- a/tests/site_upload/test_powerset_merge.py
+++ b/tests/site_upload/test_powerset_merge.py
@@ -248,9 +248,6 @@ def test_powerset_merge_single_upload(
             if study in item["Key"] and status == 200:
                 agg_df = awswrangler.s3.read_parquet(f"s3://{TEST_BUCKET}/{item['Key']}")
                 assert (agg_df["site"].eq(site)).any()
-                print(len(agg_df))
-                print(agg_df.iloc[0].to_list())
-                print(agg_df.iloc[-1].to_list())
                 assert expected_rows == len(agg_df)
                 assert first_row == agg_df.iloc[0].to_list()
                 assert last_row == agg_df.iloc[-1].to_list()


tests/site_upload/test_powerset_merge.py
```diff
@@ -209,6 +248,12 @@ def test_powerset_merge_single_upload(
             if study in item["Key"] and status == 200:
                 agg_df = awswrangler.s3.read_parquet(f"s3://{TEST_BUCKET}/{item['Key']}")
                 assert (agg_df["site"].eq(site)).any()
+                print(len(agg_df))
+                print(agg_df.iloc[0].to_list())
+                print(agg_df.iloc[-1].to_list())
+                assert expected_rows == len(agg_df)
+                assert first_row == agg_df.iloc[0].to_list()
+                assert last_row == agg_df.iloc[-1].to_list()
```
mikix, line 256: One thing that's helpful with asserts is to add data to print if it fails with a comma, so you don't need the print statements.

Like:
```
                assert expected_rows == len(agg_df), len(agg_df)
                assert first_row == agg_df.iloc[0].to_list(), agg_df.iloc[0].to_list()
                assert last_row == agg_df.iloc[-1].to_list(), agg_df.iloc[-1].to_list()
```

src/site_upload/powerset_merge/powerset_merge.py
```diff
@@ -95,10 +95,12 @@ def merge_powersets(manager: s3_manager.S3Manager) -> None:
...
             continue
+        # If the site data is being updated, don't use the last valid data
+        if manager.site in last_valid_path:
+            continue
         last_valid_metadata = functions.parse_s3_key(last_valid_path)
         subbucket_path = (
-            f"{manager.study}/{manager.data_package}/{last_valid_metadata.site}/"
-            f"{manager.study}__{manager.data_package}__{manager.version}"
+            f"{manager.study}/{manager.data_package}/{last_valid_metadata.site}/{manager.version}"
         )
```
mikix, line 104: Could more centralized path management / path class stuff help avoid this in future?

mikix (approved):

tests/site_upload/test_powerset_merge.py
dogversioning: oh i had no idea, that's incredibly useful (though i meant to strip those out)

src/site_upload/powerset_merge/powerset_merge.py
dogversioning: i suppose we could add these to the s3 manager - it already does this for aggregates, so it could do site uploads/latest/last valid instead. I'll make a ticket for this

PR #193, opened 2025-08-01
Lockfile generation

dogversioning:
This PR does the following:

- Updates some dependencies that were causing builds to fail
- Adds infrastructure to handle generating lockfiles
  - S3 manager has lockfile validation (and now has a delete command as a result
  - Fetch upload will return a lockfile id in a header, and dispatch a message to a delayed queue
  - Adds a lambda for cleaning up a lockfile after a set time period


pyproject.toml
```diff
@@ -37,12 +37,11 @@ build-backend = "flit_core.buildapi"
 
 [project.optional-dependencies]
 test = [
-    # TODO: We are investigating some new features around athena mocks, 
-    # which currently only exist in 4.1.5 and are broken in 4.1.6dev. Swing back
-    # around on this as the functionality matures.
     "coverage >=7.3.1",
     "freezegun",
-    "moto[s3,athena,sns] == 4.1.5",
```
dogversioning, line 45: queue mocks were broken in this version, which prompted the upgrade

template.yaml
```diff
@@ -52,6 +52,9 @@ Parameters:
     Type: CommaDelimitedList
   NetworkName:
     Type: String
+  LockfileDelay:
+    Type: Number
+    Default: 600
```
dogversioning, line 57: this is broken out as a parameter as a dev convenience; non-dev environments won't futz with this

template.yaml
```diff
@@ -153,14 +164,15 @@ Resources:
         LogFormat: !Ref LogFormat
         LogGroup: !Sub "/aws/lambda/CumulusAggProcessUpload-${DeployStage}-${NetworkName}"
       MemorySize: 128
-      Timeout: 800
+      Timeout: 100
```
dogversioning, line 167: just noticed this as i went along - there's not need for this value to be set so high.

tests/mock_utils.py
```diff
@@ -7,6 +7,7 @@
 TEST_PROCESS_FLAT_ARN = "arn:aws:sns:us-east-1:123456789012:test-flat"
 TEST_PROCESS_STUDY_META_ARN = "arn:aws:sns:us-east-1:123456789012:test-meta"
 TEST_CACHE_API_ARN = "arn:aws:sns:us-east-1:123456789012:test-cache"
+TEST_LOCKFILE_CLEANUP_URL = "https://sqs.us-east-1.amazonaws.com/123456789012/test-lockfile-cleanup"
```
dogversioning, line 10: these services are 95% the same thing but have completely different APIs/naming conventions. 🤷

src/shared/s3_manager.py
```diff
@@ -19,21 +23,28 @@
...
+        data_package: str | None = None,
+        version: str | None = None,
+    ):
         self.s3_bucket_name = os.environ.get("BUCKET_NAME")
         self.s3_client = boto3.client("s3")
         self.sns_client = boto3.client("sns", region_name=self.s3_client.meta.region_name)
         # If the event is an SNS type event, we're in the aggregation pipeline and set up
         # some convenience values.
-        if "Records" in event:
+        if "Records" in event and "Sns" in event["Records"][0].keys():
```
mikix, line 47: nit: don't need the `.keys()` bit

src/shared/s3_manager.py
```diff
@@ -64,6 +75,18 @@ def __init__(self, event):
...
                 f"{self.study}__{self.data_package}__{self.site}__flat.parquet"
             )
+        if study:
+            self.study = study
+        if site:
+            self.site = site
+        if data_package:
+            self.data_package = data_package
+        if version:
+            self.version = version
```
mikix, line 85: This block struck me - the  `if "Records" in event` block will set these variables too and use them to define other variables. So by setting these afterward, you now would have inconsistent state, yah? Like `self.parquet_aggregate_path` won't match these variables if they are passed in.

Would it make sense to move these above and then also drop the conditionals, so you just set these class properties from the args unconditionally, and then maybe the even overrides them? Or is the current ordering intentional?

src/shared/s3_manager.py
```diff
@@ -64,6 +75,18 @@ def __init__(self, event):
...
+        if site and study:
+            self.lockfile = (
+                f"{enums.BucketPath.META.value}/lockfiles/{self.site}__{self.study}.json"
+            )
```
mikix, line 89: This pattern here (and above) is setting a trap for yourself, in my mind - I think you want to unconditionally set `self.lockfile` to *something*, even if just `None`. Right now if you aren't setting it, it can be an attribute error to reference `self.lockfile`, and all code that does has to be careful.

Maybe something as dumb as:
```python
        self.lockfile = (
            f"{enums.BucketPath.META.value}/lockfiles/{self.site}__{self.study}.json"
        ) if site and study else None
```

Or just do `self.lockfile = None` before the if block.

src/shared/s3_manager.py
```diff
@@ -124,6 +147,16 @@ def move_file(self, from_path: str, to_path: str) -> None:
         to_path = functions.get_s3_key_from_path(to_path)
         functions.move_s3_file(self.s3_client, self.s3_bucket_name, from_path, to_path)
 
+    def delete_file(self, path: str) -> None:
+        """Moves file from one location to another in s3
```
mikix, line 151: This docstring needs updating (and to drop the to_path keyword below)

src/shared/s3_manager.py
```diff
@@ -238,3 +271,29 @@ def write_local_metadata(self, metadata: dict | None = None, meta_type: str | No
...
+
+    # Lockfile management
+    def request_or_validate_lock(self, lock_id: str | None = None):
+        try:
+            lock = functions.get_s3_json_as_dict(bucket=self.s3_bucket_name, key=self.lockfile)
+            if lock_id != lock["id"]:
+                raise errors.AggregatorStudyProcessingError
+
+        except botocore.exceptions.ClientError:
+            # if the lockfile doesn't exist, we'll make one
```
mikix, line 283: Should this maybe error out instead of making a new lock, if `lock_id` is passed in? i.e. currently, the client may think it's mid-transaction, but the server disagrees for some reason and makes a new lock, then suddenly the client is passed back a new lock unexpectedly (degenerate case, but still)

src/shared/s3_manager.py
```diff
@@ -238,3 +271,29 @@ def write_local_metadata(self, metadata: dict | None = None, meta_type: str | No
...
+            lock_id = str(uuid.uuid4())
+            lock = {"id": lock_id, "uploaded_at": datetime.datetime.now(datetime.UTC)}
+            self.s3_client.put_object(
```
mikix, line 286: If you set [If-None-Match: *](https://docs.aws.amazon.com/AmazonS3/latest/API/API_PutObject.html#API_PutObject_RequestSyntax) as a header here, I think you avoid some race condition issues.

src/shared/s3_manager.py
```diff
@@ -238,3 +271,29 @@ def write_local_metadata(self, metadata: dict | None = None, meta_type: str | No
...
+                Bucket=self.s3_bucket_name,
+                Key=self.lockfile,
+                Body=json.dumps(lock, default=str, indent=2),
+            )
+            sqs_client = boto3.client("sqs")
+            sqs_client.send_message(
+                QueueUrl=os.environ.get("QUEUE_LOCKFILE_CLEANUP"),
+                MessageBody=json.dumps({"site": self.site, "study": self.study}),
+            )
```
mikix, line 295: Might be worth sending this off before writing the lockfile, so that if you get interrupted between these two steps, you don't have a lingering lockfile.

tests/shared/test_s3_manager.py
```diff
@@ -58,6 +63,19 @@ def test_init_manager(mock_bucket):
         f"__{mock_utils.EXISTING_SITE}__{mock_utils.EXISTING_VERSION}/"
         f"{mock_utils.EXISTING_STUDY}__encounter__{mock_utils.EXISTING_SITE}__flat.parquet"
     )
+    assert manager.parquet_flat_key == (
+        f"flat/study/{mock_utils.EXISTING_SITE}/{mock_utils.EXISTING_STUDY}__{mock_utils.EXISTING_DATA_P}"
+        f"__{mock_utils.EXISTING_SITE}__{mock_utils.EXISTING_VERSION}/"
+        f"{mock_utils.EXISTING_STUDY}__encounter__{mock_utils.EXISTING_SITE}__flat.parquet"
+    )
+    assert manager.lockfile == expected_lockfile
+    manager = s3_manager.S3Manager(
```
mikix, line 72: You never test anything with this manager, except that it doesn't blow up during init.

src/shared/s3_manager.py
mikix: Well the error two lines above is handling the case of "lock file exists and it doesn't match what was provided".

The error case I'm worried about is "a lock ID was provided (i.e. the client thinks it's mid-transaction and gave the server a lock ID) but the lockfile doesn't exist". That case wouldn't hit the error two lines up.

I feel like in that case, you want to error out rather than create a new lock file - because the client thinks something else is going on - it thinks it's mid transaction.

A degenerate case, but still.

PR #194, opened 2025-08-11
Unpack study zip files

dogversioning:
This adds a new subbucket path, `staging`, in which uploaded zips are now dropped off. These are then unzipped and placed into the current existing `site_upload` subbucket.

Right now, this makes no other changes to the processing logic - that will come in a separate PR, to limit review complexity/allow us to deploy bugfixes now if we had to.


src/site_upload/unzip_upload/unzip_upload.py
```diff
@@ -0,0 +1,60 @@
...
+                .replace(f"{enums.BucketPath.STAGING.value}/", f"{enums.BucketPath.UPLOAD.value}/")
+                .replace(f"/{metadata.study}/", f"/{metadata.study}/{file.split('.')[0]}/")
+            )
+            s3_client.upload_fileobj(
+                archive.open(file), Bucket=s3_bucket_name, Key=f"{target_folder}/{file}"
+            )
+    archive_key = s3_key.replace(
+        f"{enums.BucketPath.STAGING.value}/", f"{enums.BucketPath.ARCHIVE.value}/"
+    )
+    functions.move_s3_file(
```
mikix, line 42: nit: maybe worth archiving this first and then reading from the archived version? Just so that there isn't an (admittedly unlikely) race condition between the manifest being written and this source file being overwritten by the next upload (when locking enforces a queue)

src/site_upload/unzip_upload/unzip_upload.py
```diff
@@ -0,0 +1,60 @@
...
+        s3_client=s3_client,
+        s3_bucket_name=s3_bucket_name,
+        old_key=s3_key,
+        new_key=f"{archive_key}.{datetime.datetime.now(datetime.UTC)}",
```
mikix, line 46: nit: maybe call isoformat()? I'm used to the version without the spaces, but I guess avoiding spaces is big 90's energy.

src/site_upload/unzip_upload/unzip_upload.py
mikix: But isn't it also nice to be able to debug from a canonical location, and not a location that could have been overridden since the event happened? Like, it makes sense to my brain to prioritize getting the "incoming landing zone, will be overriden" path into a "canonical, never-changing" path for processing at leisure.

But this is not a big deal and if you think locking solves this and prefer debugging without moving early, then let's keep it as it is.

dogversioning: Compromise position: I'm going to leave the way this is now, but created #195 to cover a better solution for detecting errors, which would obviate my initial concern and make this easier to reason over.

PR #196, opened 2025-09-17
Switch metadata updates to queues

dogversioning:
This PR does a pretty comprehensive overhaul of the messaging in the aggregator, to reduce race conditions and check the batch processing of uploads.

It's broken up into four separate commits, if that helps:
- Commit 1 adds a new FIFO SQS queue for metadata, pushes more of the 'what do I do with this metadata' logic down to the S3Manager where it's used, and switches stages before the S3Manager is relevant to SQS from SNS for metadata processing. It also cuts over some tests that were not using google style imports, which adds a bit of noise
- Commit 2 adds some infrastructure for interacting with the upload transaction, and adds a new lambda that reads from that queue as data appears via an event source mapping, and writes to the metadata sources in batches as the sole source of updates.
- Commit 3 adds a new lambda that uses the transaction.json to check for completeness, and triggers a glue crawl if the run is complete and another run has not already started that would be expected to cover it. The API cache has been updated to receive events from an EventBridge notification, as well as from an SNS queue.
- Commit 4 updates the integration tests, and does a bit of sanding over the changes above as a result to make sure that things are properly talking to each other


src/site_upload/update_metadata/update_metadata.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+def update_source(metadata, update):
+    for update_key, update_val in update.items():
+        if isinstance(update_val, dict):
+            metadata[update_key] = update_source(
+                metadata[update_key] if update_key in metadata.keys() else {}, update[update_key]
+            )
+        else:
+            if not (
```
mikix, line 28: nit 1: `elif not ...`

nit 2: I would remove the `not` and flip the conditional logic:
```
  update_val is not None
  or update_key not in metadata
  or metadata[update_key] is None
```
It's easier to reason about for my brain, anyway.

nit 3: you could reduce the clauses by one if you did:
```
  update_val is not None
  or metadata.get(update_key) is None
```

src/site_upload/update_metadata/update_metadata.py
```diff
@@ -0,0 +1,63 @@
...
+                metadata[update_key] = update_val
+    return metadata
+
+
+def process_event_queue(records):
+    sources = {}
+    metadata = {}
+    for record in records:
+        message = json.loads(record["body"])
+        if not any(x == message["key"] for x in sources.keys()):
```
mikix, line 43: nit: you can replace this with `message["key"] not in sources` yeah? Actually... you can replace these two lines with: `sources.setdefault(message["key"], [])` I think (and save the return value if you like, to later append to)

src/site_upload/update_metadata/update_metadata.py
```diff
@@ -0,0 +1,63 @@
...
+            sources[message["key"]] = []
+        updates = json.loads(message["updates"])
+        sources[message["key"]].append(updates)
+
+    for key, updates in sources.items():
```
mikix, line 48: So nothing else is gonna run while this is up, because of the FIFO queue stuff, yeah? i.e. we don't have to worry about two of this lambda ever running at the same time, and no other bits of code are writing to these metadata paths?

src/site_upload/cache_api/cache_api.py
```diff
@@ -83,7 +83,11 @@ def cache_api_handler(event, context):
     s3_bucket_name = os.environ.get("BUCKET_NAME")
     s3_client = boto3.client("s3")
     db = os.environ.get("GLUE_DB_NAME")
-    target = event["Records"][0]["Sns"]["Subject"]
+    if "Records" in event:
+        target = event["Records"][0]["Sns"]["Subject"]
+    elif "detail-type" in event and event["detail-type"] == "Glue Crawler State Change":
```
mikix, line 88: nit: could be `event.get("detail-type") == "Glue Crawler State Change"`

src/site_upload/check_if_complete/check_if_complete.py
```diff
@@ -0,0 +1,140 @@
...
+def check_if_complete(message) -> (bool, dict):
+    transaction = functions.get_s3_json_as_dict(
+        bucket=os.environ.get("BUCKET_NAME"),
+        key=(
+            f"{enums.BucketPath.META.value}/transactions/{message['site']}__{message['study']}.json"
+        ),
+        s3_client=s3_client,
+    )
+    upload_time = datetime.datetime.fromisoformat(transaction["uploaded_at"])
+    for filetype, source, suffix in [
```
mikix, line 29: Hmm, I guess same issue here around "forgetting to update a list of types" - maybe it's just something to live with. Or maybe there could be a central list of types? I dunno if it's worth optimizing around a forgetful future person or not. That person could also just search for `annotated_cube` and find all the spots.

src/site_upload/check_if_complete/check_if_complete.py
```diff
@@ -0,0 +1,140 @@
...
+        ("cube", enums.BucketPath.AGGREGATE.value, "aggregate"),
+        ("annotated_cube", enums.BucketPath.AGGREGATE.value, "aggregate"),
+        ("flat", enums.BucketPath.FLAT.value, "flat"),
+    ]:
+        filenames = transaction.get(filetype, [])
+        for filename in filenames:
+            dp = filename.split("__")[1].split(".")[0]
+            if source == enums.BucketPath.FLAT.value:
```
mikix, line 37: Don't you have some centralized S3 key code to handle these oddities for you?

src/site_upload/check_if_complete/check_if_complete.py
```diff
@@ -0,0 +1,140 @@
...
+            try:
+                head = s3_client.head_object(Bucket=os.environ.get("BUCKET_NAME"), Key=key)
+            except s3_client.exceptions.ClientError:
+                return False, None
+            delta = head["LastModified"] - upload_time
+            # an average time to process from an upload to all the files being
+            # processed is a minute, and since this is largely parallel it shouldn't
+            # change by much; we'll set a large buffer just in case, which should
+            # still exclude dangling files from error states
+            if upload_time > head["LastModified"] or (delta.seconds >= 300 and delta.seconds < 0):
```
mikix, line 56: 1. You made the mistake everyone makes when dealing with timedelta objects - you want to look at `total_seconds()`, not `seconds` -- `timedelta(days=1).seconds == 0` but `timedelta(days=1).total_seconds() == 86400.0)`.
2. If upload time is newer, delta.total_seconds() is negative, yeah? So you don't need the first part of this check.
3. You have an `and` in the middle of the delta.seconds check that should never be true, yeah?
4. Python has a cool feature where you can do a triple comparison, which might simplify this logic: `if not (0 <= delta.total_seconds() < 300):`
5. This should have a test if it doesn't yet :smile:

src/site_upload/check_if_complete/check_if_complete.py
```diff
@@ -0,0 +1,140 @@
...
+
+
+def has_new_packages(message, transaction) -> bool:
+    db = os.environ.get("GLUE_DB_NAME")
+    s3_bucket_name = os.environ.get("BUCKET_NAME")
+    tables = awswrangler.athena.read_sql_query(
+        (
+            f"SELECT table_name FROM information_schema.tables "  # noqa: S608
+            f"WHERE table_schema = '{db}' "  # nosec
+            f"AND regexp_like(table_name, '^{message['study']}__')"
```
mikix, line 68: Two different ways of turning off the S608 check? And not needed on the third line? That third line actually does seem like a risk - that's string is straight from the network, yeah? Maybe some validation that its characters are alphanumeric-plus-underscore.

src/site_upload/check_if_complete/check_if_complete.py
```diff
@@ -0,0 +1,140 @@
...
+        database=db,
+        s3_output=f"s3://{s3_bucket_name}/awswrangler",
+        workgroup=os.environ.get("WORKGROUP_NAME"),
+    )["table_name"].values.tolist()
+    for dp_type in ["cube", "annotated_cube", "flat"]:
+        filenames = transaction.get(dp_type, [])
+
+        for filename in filenames:
+            dp = filename.split("__")[1].split(".")[0]
+            if dp_type == "flat":
```
mikix, line 79: This branching logic on flat feels very similar to previous code - worth centralizing probably, as these keys are so specific and easy to mess up, if you're not copying/pasting - and if you're doing that, it should be centralized :smile:

src/site_upload/check_if_complete/check_if_complete.py
```diff
@@ -0,0 +1,140 @@
...
+
+
+def cleanup_transaction(message) -> bool:
+    key = f"{enums.BucketPath.META.value}/transactions/{message['site']}__{message['study']}.json"
+    try:
+        s3_client.head_object(Bucket=os.environ.get("BUCKET_NAME"), Key=key)
+    except botocore.exceptions.ClientError:
+        return False
+    s3_client.delete_object(Bucket=os.environ.get("BUCKET_NAME"), Key=key)
+    return True
```
mikix, line 95: I'd say just skip the head and directly do the delete, handling "not there" errors. Otherwise there's a tiny race condition where the file can get created between these two - probably doesn't matter here, but just bumped me.

src/site_upload/check_if_complete/check_if_complete.py
```diff
@@ -0,0 +1,140 @@
...
+                if "LastCrawl" in crawler and crawler["LastCrawl"][
+                    "StartTime"
+                ] > datetime.datetime.fromisoformat(event["Records"][0]["Sns"]["Timestamp"]):
+                    return functions.http_response(
+                        200, f"Recrawl request for {message!s} covered by subsequent request"
+                    )
+                g_client.start_crawler(Name=os.environ.get("GLUE_CRAWLER_NAME"))
+                return functions.http_response(200, f"Crawl for {message!s} initiated")
+            attempts += 1
+            sleep(60)
```
mikix, line 126: Won't a caller to this lambda time out over the course of ten minutes? Maybe the ten-retries-then-give-up logic should be client-side?

src/site_upload/check_if_complete/check_if_complete.py
```diff
@@ -0,0 +1,140 @@
...
+            crawler = g_client.get_crawler(Name=os.environ.get("GLUE_CRAWLER_NAME"))["Crawler"]
+            if crawler["State"] == "READY":
+                cleanup_performed = cleanup_transaction(message)
+                if not cleanup_performed:
+                    return functions.http_response(
+                        200, f"Processing request for {message!s} already sent"
+                    )
+                if "LastCrawl" in crawler and crawler["LastCrawl"][
+                    "StartTime"
+                ] > datetime.datetime.fromisoformat(event["Records"][0]["Sns"]["Timestamp"]):
```
mikix, line 119: This got formatted real weird by ruff - maybe use some variables to keep the line length down.

src/site_upload/check_if_complete/check_if_complete.py
```diff
@@ -0,0 +1,140 @@
...
+                    return functions.http_response(
+                        200, f"Recrawl request for {message!s} covered by subsequent request"
+                    )
+                g_client.start_crawler(Name=os.environ.get("GLUE_CRAWLER_NAME"))
+                return functions.http_response(200, f"Crawl for {message!s} initiated")
```
mikix, line 124: Will this start a crawler twice? Like... where's the happy path that exits the "start a crawler and wait for result" loop and does the `TOPIC_CACHE_API_ARN` stuff at the bottom?

pyproject.toml
```diff
@@ -45,6 +45,7 @@ test = [
     "pytest-mock",
     "pytest-subprocess",
     "responses",
+    "toml"
```
mikix, line 48: You could always just bump to 3.11 :smile:

src/shared/functions.py
```diff
@@ -58,7 +70,10 @@ def http_response(
     if extra_headers:
         headers.update(extra_headers)
     if status >= 200 and status < 300:
-        logging.info(body)
+        if alt_log:
+            logging.info(alt_log)
+        else:
+            logging.info(body)
```
mikix, line 76: nit: Mike's one liner obsession: `logging.info(alt_log or body)`

mikix (approved):

src/site_upload/check_if_complete/check_if_complete.py
dogversioning: not yet - i can go the other way. I'm going to add this soon, because this path munging is starting to give me a headache.

src/site_upload/check_if_complete/check_if_complete.py
dogversioning: yeah, agree - was going to do it as a followup due to the size of this.

src/site_upload/check_if_complete/check_if_complete.py
```diff
@@ -0,0 +1,140 @@
...
+    return True
+
+
+# @decorators.generic_error_handler(msg="Error processing metadata events")
+def check_if_complete_handler(event, context):
+    del context
+    message = json.loads(event["Records"][0]["Sns"]["Message"])
+    completed, transaction = check_if_complete(message)
+    if not completed:
+        return functions.http_response(200, "Processing not completed")
```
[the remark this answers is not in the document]
dogversioning: Can't hurt

src/site_upload/check_if_complete/check_if_complete.py
dogversioning: So - this should cause this to return early, and then the crawl will kick off, and the notification for the next step comes from the eventbridge detection of the crawl ending? Am I missing something?

src/site_upload/check_if_complete/check_if_complete.py
dogversioning: 15 minutes is the lambda timeout - this should be safe.

src/site_upload/process_upload/process_upload.py
```diff
@@ -68,21 +69,22 @@ def process_upload(s3_client, sns_client, sqs_client, s3_bucket_name: str, s3_ke
...
         functions.move_s3_file(s3_client, s3_bucket_name, s3_key, new_key)
+        print(f"{data_package}__{version}")
         metadata = functions.update_metadata(
             metadata={},
             site=site,
             study=study,
-            data_package=data_package,
-            version=version,
+            data_package=data_package.split("__")[1],
+            version=f"{data_package}__{version}",
```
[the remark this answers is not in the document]
dogversioning: once i can go both directions with key <-> metadata, I think a lot of this can get cleaned up.

src/site_upload/unzip_upload/unzip_upload.py
```diff
@@ -22,6 +22,20 @@ def unzip_upload(s3_client, sns_client, s3_bucket_name: str, s3_key: str) -> Non
     files = archive.namelist()
     files.remove("manifest.toml")
 
+    # We'll update the transaction data with the files we're going to process
+    # (we can't use the manifest, because empty tables will not get uploaded),
+    # and use this later to check to see if all files have been processed.
+    # Since metadata is just copied and not otherwise massaged, we skip it
+    manager = s3_manager.S3Manager(site=metadata.site, study=metadata.study)
+    transaction = manager.get_transaction()
+    for upload_type in [
```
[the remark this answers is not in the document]
dogversioning: Right now, metadata isn't captured by this - this is just the ones that go through a more complicated archiving process.

src/site_upload/update_metadata/update_metadata.py
dogversioning: Correct - this is a property of subscribing a lambda to a FIFO queue.

src/site_upload/update_metadata/update_metadata.py
dogversioning: First one is fine - setdefault i'm gonna sit on, that's something i don't keep in the front of the brain.

src/site_upload/check_if_complete/check_if_complete.py
mikix: I think my brain skipped a beat and didn't see the return. Ignore my above comment :smile:

src/site_upload/check_if_complete/check_if_complete.py
dogversioning: Yeah, these are almost all internal HTTP calls for tracking success/failure status.

pyproject.toml
dogversioning: the stdlib tomllib is read only, not write 🤷

PR #203, opened 2025-12-05
Added more S3 state management scripts

dogversioning:
This adds two new scripts for managing artifacts in S3.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -9,7 +9,7 @@ dependencies= [
     "arrow >=1.2.3",
     "awswrangler >=3.5, <4",
     "boto3",
-    "cumulus-library <5",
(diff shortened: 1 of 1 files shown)


template.yaml
```diff
@@ -417,6 +417,11 @@ Resources:
       Environment:
         Variables:
           BUCKET_NAME: !Sub '${BucketNameParameter}-${AWS::AccountId}-${DeployStage}-${NetworkName}'
+      Events:
+        MetadataUpdateSQSEvent:
+          Type: SQS
+          Properties:
+            Queue: !GetAtt SQSMetadataUpdate.Arn
```
dogversioning: This was unrelated but ports over a re-insert from the deployment repo.

PR #217, opened 2026-04-09
Updated for library v6

dogversioning:
This updates the aggregator to use V6 of the library. As a byproduct of this change, it swaps out `freeze_time` for `time_machine`, to avoid issues mocking the datetime class directly.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "cumulus-aggregator"
-requires-python = ">= 3.10"
+requires-python = ">= 3.11"
 version = "0.3.0"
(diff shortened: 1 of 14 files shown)


pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [project]
 name = "cumulus-aggregator"
-requires-python = ">= 3.10"
+requires-python = ">= 3.11"
```
dogversioning, line 3: this has been wrong for... awhile, but only affects running tests locally, the SAM config stuff handles the versioning of this on deployment.

mikix (approved):

PR #224, opened 2026-08-15
Support VERSION = 000 is a dev version

msa2984:
Resolves #212 

### Supported / Tested scenarios:
Updated to match the expected behavior better. 
| Description | Outcome |
| ----------- | --------- |
| Uploading 000 as your very first upload | Upload is successful, 000 metadata added |
| 005 -> 000 | 000 is uploaded after 005, no difference in behavior |
| 005 -> 006 -> 000 | 000 is uploaded, no change in 005 or 006 versions |
| 000 -> 005 -> 006 -> 000 | second instance of 000 triggers crawler as expected, old 000 files are removed / updated |
| 000 -> 005 -> 000, schema changed | second instance of 000 triggers crawler as expected, old 000 files are removed / updated |

### Notes
- The DPU usage for the crawler is higher in the `RESERVED_DEV_VERSION`, likely due to the extra processing to iterate through the metadata to remove old data.
- Unfortunately, it seems that there isn't an easy way to remove the old information from the `ARCHIVE` folder. Perhaps there should be an extra event triggered for this cleanup after the fact? It causes a lot of errors when I try to handle it at the same time as the other 'bulk deletion' for aggregates, etc.


src/site_upload/powerset_merge/powerset_merge.py
```diff
@@ -89,8 +101,27 @@ def merge_powersets(manager: s3_manager.S3Manager) -> None:
     logger.info(f"Proccessing data package at {manager.s3_key}")
     # initializing this early in case an empty file causes us to never set it
     df = pandas.DataFrame()
+
+    if manager.version == consts.RESERVED_DEV_VERSION:
+        functions.remove_previous_uploads(
```
msa2984: I'm not sure if there's a better place to do this sort of deletion. It seems to go against the standard set for updating metadata (i.e. emitting an event and merging in the updates), however, the old data needs to be removed prior to merging the data frames, otherwise we will continue to merge in the old data that the dev version is supposed to be removing. This seemed to be the earliest time where the old files could be removed without causing failures in the upload step.

src/site_upload/update_metadata/update_metadata.py
```diff
@@ -27,6 +27,38 @@ def update_source(metadata, update):
     return metadata
 
 
+def remove_stale_study_metadata(node: dict, study: str, version: str) -> None:
```
msa2984, line 30: This is actually the bulk of the change for this PR - 

The current behavior of update_metadata is additive - i.e., it is always assumed that we are merging in new metadata content, and never removing old metadata content. To preserve the current functionality, I've added some extra info to the message passed to the `update_metadata_handler`, namely `version` and `study`, so that if we detect `version == RESERVED_DEV_VERSION`, we recursively check the current metadata file for any keys that match the `study` but do NOT match the `version`, they are removed from the metadata file. 

Originally I wanted to just set the metadata collection to {} when we detect this scenario instead of recursing through the file since this is generally a slow operation, however for some metadata events this can actually unintentionally remove some of the `000` entries.

dogversioning:
> ### Supported / Tested scenarios:
> 005 -> 000 	000 removes 005 artifacts, crawler is triggered, 005 tables removed, 005 metadata removed
> 005 -> 006 -> 000 	000 removes 005, 006 artifacts, crawler is triggered, 005 and 006 tables removed, 005 and 006 metadata removed

I think this is not the intended behavior - in both cases, I would expect the 000 version to be removed, but 005/006 to stay, depending on context, so that you don't blow away good data with dev data

>     * Unfortunately, it seems that there isn't an easy way to remove the old information from the `ARCHIVE` folder. Perhaps there should be an extra event triggered for this cleanup after the fact? It causes a lot of errors when I try to handle it at the same time as the other 'bulk deletion' for aggregates, etc.

It was requested that we basically don't touch this data, so that if something strange happened, we could use this as a way to go back and investigated what happpened. To date we never had to do this. But if we were going to have any kind of cleanup, it would probably be based on created date, not done ad hoc.

docs/integration.md
```diff
@@ -48,12 +48,13 @@ If the quick test was successful, you can test your processing pipeline entirely
   The setup guide in the [Cumulus ETL documentation](https://docs.smarthealthit.org/cumulus/etl/)
   includes instructions to deploy with a synthetic dataset.
 - When it's complete, you should be able to view data in athena to verify.
-- In the cumulus library repo, build the Athena tables and export results, with\
```
dogversioning, line 51: wow this comment is SO OLD, how did we never catch it

msa2984:
> > ### Supported / Tested scenarios:
> > 005 -> 000 	000 removes 005 artifacts, crawler is triggered, 005 tables removed, 005 metadata removed
> > 005 -> 006 -> 000 	000 removes 005, 006 artifacts, crawler is triggered, 005 and 006 tables removed, 005 and 006 metadata removed
> 
> I think this is not the intended behavior - in both cases, I would expect the 000 version to be removed, but 005/006 to stay, depending on context, so that you don't blow away good data with dev data
> 
> > ```
> > * Unfortunately, it seems that there isn't an easy way to remove the old information from the `ARCHIVE` folder. Perhaps there should be an extra event triggered for this cleanup after the fact? It causes a lot of errors when I try to handle it at the same time as the other 'bulk deletion' for aggregates, etc.
> > ```
> 
> It was requested that we basically don't touch this data, so that if something strange happened, we could use this as a way to go back and investigated what happpened. To date we never had to do this. But if we were going to have any kind of cleanup, it would probably be based on created date, not done ad hoc.

OH - I see, we only want to override the old 000 data. That's a lot easier than what I've done here, I'll update the PR

src/site_upload/powerset_merge/powerset_merge.py
dogversioning: You could? Do it once at the front after the processing upload step before you start moving data around? or maybe even in front with a new event target so you could have a different flow? i realize that's a big structural change so maybe not worth it at this point.

dogversioning (commented):
I think this generally looks good as is, other than the spec being up for discussion - i'm hoping this might be as easy as flipping around the polarity on remove_stale_metadata? But pausing on the approval while we has that bit out.

dogversioning (approved):

src/shared/functions.py
```diff
@@ -198,12 +200,37 @@ def write_metadata(
...
+    for prefix in (
+        enums.BucketPath.AGGREGATE,
+        enums.BucketPath.FLAT,
+        enums.BucketPath.LAST_VALID,
+        enums.BucketPath.MANIFEST,
+        enums.BucketPath.STUDY_META,
+    ):
+        for key in get_s3_keys(s3_client, s3_bucket_name, prefix):
+            parsed_key = parse_s3_key(key)
+            if parsed_key.version == version and parsed_key.study == study:
```
msa2984, line 230: changed this logic to only delete for the old DEV instance, not the non-DEV versions

src/site_upload/powerset_merge/powerset_merge.py
msa2984: okay - I've moved it to unzip_upload... this seems more appropriate? I thought about making a new event target too, but I was thinking that's a lot of change for what is ultimately supposed to be a dev convenience

msa2984:
okay @dogversioning , I've changed the logic for the metadata and file upload steps so that we are pruning the data from old 000 uploads. I think there's still a question around deleting the standalone files, so I don't want to merge until you've had another chance to review

